PR #6, opened 2015-07-23
Update aliases.md

jtreuting:
The GetNumber MethodAlias didn't match the GetMagicNumber call in the sample code below


patriksvensson:
Good catch! Thanks! :+1:

PR #11, opened 2015-10-08
Explain how to abort the build

mgnslndh:
I used Patriks answer as a base along with my example from my question on StackOverflow. http://stackoverflow.com/q/33014431/966


patriksvensson:
Excellent initiative :smile: 
Will merge this later tonight when I get home from work :+1:

PR #16, opened 2015-10-20
Update installing-tools.md

pheiberg:
Correct spelling


patriksvensson:
Good catch! I always get `disk` wrong for some reason :smile:

PR #31, opened 2015-11-19
Blog Post about VSO and VS Code Extensions

gep13:


gep13:
@patriksvensson @devlead let me know what you think about this blog post.

In order for it to work properly, you will need to first pull in this https://github.com/cake-build/cake-vscode/pull/1 otherwise the images will not work.

It was sort of rushed, so feel free to suggest changes.  I just thought it would be worth getting this out there.

PR #62, opened 2016-02-15
Update and rename criterias.md to criteria.md

markrendle:
"Criteria" is the plural form of "criterion". "Criterion" is an ugly word, but not as ugly as "criterias". Technically, a single predicate could involve multiple criteria, so I've just removed the "s".


gep13:
@markrendle thanks for the correction :+1:

PR #67, opened 2016-03-02
adds Cake.Npm and Cake.Gulp addins

Philo:


devlead:
@Philo Hi, namespaces need to be documented you can see an example of that here:
https://github.com/WCOMAB/Cake.Slack/blob/master/src/Cake.Slack/Properties/Namespaces.cs

Philo:
@devlead - I've added the class as above to both Cake.Npm and Cake.Gulp

devlead:
Ok great, will review and merge if ok later.

One tip saw that you currently check for gulp.cmd & node.exe, to be xplat you should also check for `node`, `nodejs` and `gulp` in your `GetToolExecutableNames()`

devlead:
@Philo tip 2 also your nugets don't need a dependency on Cake.Core as it's already in place when add-ins are loaded, saves a little time when add-ins are installed (otherwise Cake.Core is downloaded and installed in the add in folder but never used).

devlead:
@Philo this is now merged and available on the website, thanks for your contribution :+1:

Philo:
@devlead, thanks for the feedback, will update accordingly.  I was hoping someone  wouldpoint me towards to xplat locations for node and nom so  Ican include xplat support

PR #73, opened 2016-04-06
Update features.md

RLittlesII:
Added additional supported build systems and SpecFlow


gep13:
Thanks for updating this! :+1:   There are actually a few more things that we need to add into this list, will have to get round to adding them :smile:

PR #78, opened 2016-05-03
Extending Bootstrapper Documentation

RLittlesII:
This doesn't include the bash.  I just wanted to be sure this was useful before I went any further. I'll rebase before any merge.

### DIFF
--- a/src/Cake.Web/App_Data/docs/tutorials/extending-the-bootstrapper.md
+++ b/src/Cake.Web/App_Data/docs/tutorials/extending-the-bootstrapper.md
@@ -1,0 +1,69 @@
+#Extending the Cake Bootstrapper
+
+The Cake Bootstrapper that you can get directly from [cakebuild.net](http://cakebuild.net) is intended as a starting point for what can be done. It is the developer's discretion to extend the bootstrapper to solve for your own requirements.
+
+##Adding an additional parameter
+ For example, let's say you want to add an input parameter to the bootstrapper for mobile build configurations.  These are the changes that you could make in order to achieve this...
+
+- Add parameter to the parameter list
+
+####PowerShell
+```powershell
+Param(
+    [string]$Script = "build.cake",
+    [string]$Target = "Default",
+    [ValidateSet("Release", "Debug")]
+    [string]$Configuration = "Release",
+    [ValidateSet("Debug", "Release", "AdHoc", "AppStore")]
+    [string]$BuildConfiguration = "AdHoc", # Added parameter
+    [ValidateSet("Quiet", "Minimal", "Normal", "Verbose", "Diagnostic")]
+    [string]$Verbosity = "Verbose",
+    [switch]$Experimental,
+    [switch]$WhatIf,
+    [switch]$Mono,
+    [switch]$SkipToolPackageRestore,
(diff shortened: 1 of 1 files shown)


gep13:
@RLittlesII I think that this makes sense to me, yes.  In the heading section, would it be possible to call out what you are doing?  i.e. give it a scenario.

`The Cake Bootstrapper that you can get directly from cakebuild.net is intended as a sample on what can be done.  It is entirely possible that you can extend the bootstrapper to include your own requirements.  For example, let's say you want to add you own input parameter to the bootstrapper called buildConfiguration.  These are the changes that you would need to make in order to achieve this...`

What do you think?

RLittlesII:
I like it.  I have another scenario to add, so that definitely makes sense.  I'll update with the rest of the documentation and scenarios. Is it necessary to add a bash example?

gep13:
@RLittlesII for the sake of completeness, I would say that yes, we should have the same examples in bash, as well as PowerShell.

RLittlesII:
@gep13 I'll make sure to prefix it to show PowerShell vs Bash.

RLittlesII:
@gep13 I still have to test that the documentation works on bash.  Comments on making sure the look and feel of the document vibe with the rest of the cake site is appreciated.  I don't want this to read like it came from me.  :smile:

gep13:
@RLittlesII this looks really good to me so far!  The only comment would be to perhaps make the changes that you have made in the code snippets bold, so that they can stand out.  What do you think?

RLittlesII:
@gep13 I think calling out the specific change would ultimately make it more useful.  I am not aware of how to bold something inside of a code block in mark down.  Maybe I add a comment on the end of the line that says `// This line was added` or something to that affect?  If you know how, let me know.

gep13:
```
**bold**
not bold
```

gep13:
@RLittlesII hmm, I thought there was a direct way, but looks like there isn't :-(

In the above, I did what was suggested here:

http://meta.stackexchange.com/questions/183610/how-to-combine-bold-and-code-sample-in-markdown

gep13:
But anyway, I agree, calling out the specific thing that has changed, would ultimately make the most sense.

RLittlesII:
@gep13 I found the same post and was equally disappointed that it did not work.  I will add comments to the code sample.

RLittlesII:
@gep13 Are the changes sufficient?  I have tried to bold the line, but it doesn't seem to work correctly.

gep13:
@RLittlesII this all looks good to me, with only one suggestion...

Can you add an example of the command that you would use to run the bootstrapper in each case?  This would be especially helpful in the second example, to show passing in two build configurations.

Thanks!

RLittlesII:
@gep13 I just realized that passing array's at the bash shell is clunky.  I need to prove it works before I add it.

gep13:
@RLittlesII not a problem at all.  Let me know when you are ready for me to take a look again.

RLittlesII:
@gep13 I am inclined to remove the part about array's that way we can close this out and I can add it once I've done the research.

gep13:
@RLittlesII that sounds fine with me.  You want to fix it up and I can get it merged in?

RLittlesII:
Fixed up and rebased.

gep13:
@RLittlesII perfect, thank you!

Issue #82, opened 2016-05-11
Add section for VSO


patriksvensson:
We currently have information about [VSO](http://cakebuild.net/docs/build-systems/vso) that is hidden. Add a new section (Build Systems) and add this page here.

patriksvensson:
Closed by #90.

PR #90, opened 2016-05-26
Added Build Systems Section

gep13:


gep13:
@patriksvensson @devlead looks like the Build System section was already there.  Just needed to be exposed :smile:

src/Cake.Web/App_Data/docs.xml
```diff
@@ -35,7 +35,7 @@
     <item title="Visual Studio Code" file="editors/vscode.md" />
   </section>
   <section id="build-systems" title="Build Systems">
-    <item title="Visual Studio Online" file="build-systems/vso.md" hidden="true"/>
```
patriksvensson: Not sure we can change the URI of this one since we link to it from the marketplace.

gep13: @patriksvensson ah, good catch! I have changed this back.

patriksvensson: There is support in the site for legacy redirects. Maybe you can add it there and just redirect it to the new URI?

patriksvensson: Have a look here: https://github.com/cake-build/website/blob/develop/src/Cake.Web/App_Start/LegacyRouteConfig.cs

gep13: I don't think it is a big deal.  I was just being pedantic, and thought I would change it while I was there.  I am happy to leave it as is.

patriksvensson: @gep13 Absolutely!

patriksvensson:
Look good to me. Going to merge this then! :+1:

patriksvensson:
Related to #82.

Issue #92, opened 2016-06-06
Contributing page is outdated


endrton:
The [contributing page](http://cakebuild.net/docs/contributing/guidelines) shall be synced with the info on the [github contributing page](https://github.com/cake-build/cake/blob/develop/CONTRIBUTING.md)

gep13:
This has now been corrected here:

http://cakebuild.net/docs/contributing/guidelines

PR #93, opened 2016-06-06
Added support for mirroring documents.

patriksvensson:
Tested and works on my machine, but would feel better if someone else tried this out before merging.


patriksvensson:
Should take care of issue #92 submitted by @endrton.

src/Cake.Web/Cake.Web.csproj
```diff
@@ -267,7 +267,53 @@
     <Content Include="Views\Blog\Category.cshtml" />
     <Content Include="Views\Addin\Index.cshtml" />
     <Content Include="Views\Shared\_AddinWarning.cshtml" />
-    <Content Include="App_Data\**\*.md" />
+    <Content Include="App_Data\blog\2015-07-05-cake-v0.4.3-released.md" />
```
patriksvensson: Hmm, not sure what happened here. Strange.

gep13: @patriksvensson looks like Visual Studio trying to be "clever". Do you want to put this back the way it was, or do you want them explicitly added into the csproj file?

devlead: @gep13 @patriksvensson  We don't want to add each explicitly, so that part needs to be reverted.

patriksvensson:
@gep13 This should be fixed now.

gep13:
@patriksvensson this looks good to me!

PR #97, opened 2016-06-10
Reworded error-handling.md

DavidBrower:
The phrase "without affecting it's propagation or resulting stack trace" is incorrect ("its" is short for "it is" and "its" is the possessive of the pronoun "it".

I have reworded the phrase to use the possessive pronoun.


gep13:
@DavidBrower thanks for catching and fixing this! :+1:

PR #101, opened 2016-06-17
Update getting-started.md to reference xUnit instead of NUnit a second time.

eeevans:
I'm not sure if it was intended but NUnit is referenced twice and xUnit not at all.


devlead:
@eeevans  Nice catch, thanks! Merged 👍

PR #105, opened 2016-06-22
Add new Cake.NSwag addin, and update AutoRest categories

agc93:
Just another new addin, and updating an old addin to be in a new category.

Let me know if there's any issues.


gep13:
@agc93 looks good to me!

PR #107, opened 2016-06-23
added Cake.Plist by @reicheltp

ghuntley:


src/Cake.Web/App_Data/addins.xml
```diff
@@ -364,4 +364,15 @@
...
   </Addin>
+  <Addin>
+    <Name>Cake.Plist</Name>
+    <NuGet Id="Cake.Plist">
+      <Filter>/**/Cake.Plist.dll</Filter>
+      <Filter>/**/Cake.Plist.xml</Filter>
+    </NuGet>
+    <Repository>https://github.com/reicheltp/Cake.Plist</Repository>
+    <Author>Paul Reichelt</Author>
+    <Description>Cake addin for deserializing, modifying and serializing Apple macOS/iOS xml plists directly from .NET source code.</Description>
```
patriksvensson: Shouldn't `macOS` be `OS X`?

ghuntley: OSX is now macOS :trollface: :-1: (god damn it apple)

http://www.theverge.com/2016/6/22/11997918/apple-macos-sierra-preview-siri-icloud-drive

gep13: @ghuntley until that happens, I think we should stick to OS X for now, and then we can do an update across the entire site when it ships.  Thoughts?

patriksvensson: @gep13 @ghuntley Yes, I think that is a good idea.

ghuntley:
Changed to OSX instead of macOS.

gep13:
@ghuntley thank you for making the update as requested.  This should be live on the site soon.

PR #108, opened 2016-06-27
Updated to not use obsolete Setup/Teardown

devlead:
Examples use obsolete Setup/Teardown methods.


gep13:
@devlead thank you for getting this updated! :+1:

PR #112, opened 2016-07-12
Change VSO to VSTS

pascalberger:
Change Visual Studio Online to Visual Studio Team Services


gep13:
@pascalberger good catch!  I thought I had caught all of these the last time I updated this!

PR #124, opened 2016-07-21
Added PropertyType to properties-table in type.cshtml

ristaloff:
I'm not sure this is the best way to do this. But it's at least a start.


ristaloff:
#124 Added an image here, so you can see what it looks like.

ristaloff:
Sure. I should close this first then?

gep13:
Yip, I can do that though :smile:

PR #126, opened 2016-07-22
added Cake.AppleSimulator addin

ghuntley:


gep13:
Sweet! Thanks for the addition!

gep13:
@ghuntley and it is live: http://cakebuild.net/dsl/applesimulator

One thing, can you think about adding a summary tag to your aliases, similar to what we do here:

image

For the built in Aliases?  This really helps people understand what they are, and what is required to get started using them. Thanks!

gep13:
This is what I did for my recent Twitter addin:

image

Issue #131, opened 2016-07-27
Add Modules section


gep13:
With the ability to support additional Cake Modules, I think we need an additional section on the site, which allows people to submit information about custom Cake modules, similar to what they can do for Addins.

daveaglick:
Educate me - what's a module and how is it different from an addin? Is there a prototypical example module I should take a look at?

devlead:
A module differs from an addin that it instead of adding features, replaces(or extends) core functionality in Cake
An example of this is [Cake.LongPath.Module](https://github.com/cake-contrib/Cake.LongPath.Module) which replaces the filesystem abstraction with one that supports long paths on .NET 4.5.

gep13:
@daveaglick another example would be this:

https://github.com/larzw/Cake.Paket

Where the #tool and #addin pre-processor directives are being extended to include support for another package manager, in addition to nuget.

gep13:
@daveaglick I think this one would be a good candidate for a pairing session as well.

fwinkelbauer:
This documentation should also include how to handle the packages.config file in the tools/modules folder. Here's the `.gitignore` entry that I am using:

```
tools/**/*
!tools/packages.config
!tools/modules
!tools/modules/packages.config
```

Line 3 and 4 might look weird, but you seem to need both.

gep13:
@fwinkelbauer yes, that is a good idea!

pascalberger:
Mockup for replacing current addins page with an extension page which can also show modules: https://github.com/cake-build/website/issues/1096#issuecomment-711096108

pascalberger:
https://github.com/cake-build/website/pull/1522 added support for showing modules in extension list.

Last missing part is adding yml files for known modules which is tracked by https://github.com/cake-contrib/Cake.AddinDiscoverer/issues/26.

pascalberger:
Modules are now listed at https://cakebuild.net/extensions/. It's possible to filter the list to modules or addins through the filter UI or by searchging for `modules` or `addins`

PR #133, opened 2016-07-28
[WIP] GH131: Added Modules section

gep13:
Relates to #131 

Work is still required here, but wanted to put it up for review.

### DIFF
--- a/src/Cake.Web.Core/Content/Modules/Module.cs
+++ b/src/Cake.Web.Core/Content/Modules/Module.cs
@@ -1,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+using Cake.Web.Core.NuGet;
+using Cake.Web.Docs;
+
+namespace Cake.Web.Core.Content.Modules
+{
+    public sealed class Module : IDocumentationMetadata
+    {
+        public string Name { get; set; }
+        public Uri Website { get; set; }
+        public Uri Repository { get; set; }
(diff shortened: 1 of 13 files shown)


src/Cake.Web.Core/Content/Modules/Module.cs
```diff
@@ -0,0 +1,28 @@
...
+        public string Name { get; set; }
+        public Uri Website { get; set; }
+        public Uri Repository { get; set; }
+        public string Author { get; set; }
+        public string Description { get; set; }
+        public List<string> Categories { get; set; }
+        public PackageDefinition PackageDefinition { get; set; }
+
+        public Uri Url => Repository ?? Website;
+        public bool IsOwnedByAddin => true;
```
devlead: We should add an IsOwnedByModule and so we i.e. can add Modules at
https://github.com/cake-build/website/blob/97fb9986af1f233f87caeb8def3bfeaad4c58225/src/Cake.Web/Views/Dsl/Category.cshtml#L35

gep13: @devlead Interesting. I hadn't even noticed that property was in there (yet).  It was a blind copy/paste of the Addin class, just to get something working.  Will look into this.

devlead:
@gep13 think we could close this and revisit when Wyam migration is done?

gep13:
@devlead yes, I think you are probably right.

PR #139, opened 2016-08-10
Added info about installing specific versions or pre-release of tools

JakeGinnivan:


JakeGinnivan:
Ah, got pointed at http://cakebuild.net/docs/fundamentals/preprocessor-directives

Maybe should link to that instead?

gep13:
@JakeGinnivan Yes, I think a link to the Preprocessor Directive page, would make sense.

PR #148, opened 2016-09-02
Addin two new addins

gep13:
@cake-build/team ok, let's try this again...

I have tested this locally, and it now works.  I got @redth to push new versions of each nuget package, that no longer have the duplicated namespaces.


devlead:
@gep13 I'll take it for a spin.

devlead:
@gep13 Works on my machine :tm:

PR #150, opened 2016-09-05
Adds Cake.Figlet to addins list

enkafan:


gep13:
@enkafan it is now live on the site: http://cakebuild.net/dsl/figlet thanks again!

PR #152, opened 2016-09-08
Added Cake.SqlServer

trailmax:
Added a description of the new SqlServer package into list of addins

### DIFF
--- a/src/Cake.Web/App_Data/addins.xml
+++ b/src/Cake.Web/App_Data/addins.xml
@@ -521,6 +521,7 @@
     <Name>Cake.SqlServer</Name>
     <NuGet Id="Cake.SqlServer">
       <Filter>/**/Cake.SqlServer.dll</Filter>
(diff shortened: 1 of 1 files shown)


gep13:
@trailmax can I get you to do something for me?

As this PR stands right now, the generated documentation looks like this:

image

Which is missing all the documentation that you put into the source code here:

image

This is due to the fact that you are not including the XML Documentation file within in the NuGet Package.  Would you be in a position to add this file into the NuGet Package, so that the documentation can be generated?

gep13:
Raised an issue about this here: https://github.com/AMVSoftware/Cake.SqlServer/issues/3

trailmax:
Added the missing xml file into nuget and added the reference to this file in the addins.xml.

gep13:
@trailmax thanks for getting this added.  This is now live here: http://cakebuild.net/dsl/sqlserver

PR #162, opened 2016-09-24
GH134: Added resources section

gep13:


gep13:
@patriksvensson have you put your presentation deck from your recent one onto slideshare or somewhere similar?

patriksvensson:
@gep13 No, I haven't but will try to do that soon.

Otherwise this PR looks good to me! OK with merging?

gep13:
@patriksvensson yip, fine with me. We can add for items as and when we know about them.

PR #182, opened 2016-10-15
Add a page describing how to pass information using arguments and env…

xclayl:
Added a page describing how to pass information using arguments and environment variables


devlead:
@xclayl your changes have now been merged and will be avail with next publish of the site, thanks for contributing 👍

xclayl:
Thanks @devlead

devlead:
We're updating the bootstrappers and will come with new guidance as soon as that's done. Please if possible raise an separate issue about multiple args so we don't miss that.

PR #183, opened 2016-10-19
Added Cake.DoInDirectory

pitermarx:


src/Cake.Web/App_Data/addins.xml
```diff
@@ -539,4 +539,14 @@
     <Description>Cake Addin to allow message posting to HipChat.</Description>
     <Categories>Communication</Categories>
   </Addin>
+  <Addin>
+    <Name>Cake.DoInDirectory</Name>
+    <NuGet Id="Cake.DoInDirectory">
+      <Filter>/**/Cake.DoInDirectory.dll</Filter>
```
gep13: @pitermarx in order to generate the documentation on the site, we are going to need the associated XML file that can be generated as part of the build process.  This can be done by enabling that setting within the project properties.  Let me know if you are not sure how to do this.

pitermarx:
how about now?

src/Cake.Web/App_Data/addins.xml
pitermarx: done

gep13:
@pitermarx this looks like exactly what we would need.  However, ideally this would be done as 1 commit, not 3.  Since you haven't created a branch to do this change on, squashing these into a single commit becomes tricky.  Would you be able to re-do this work on a separate branch, and create another pull request?

Thanks!

pitermarx:
I was expecting you'd say that but as I didn't know how to do it without starting over I hoped it was ok like this. 
I'll do it again then

--
Pedro Marques

No dia 21/10/2016, às 08:52, Gary Ewan Park <notifications@github.com> escreveu:

gep13:
>@pitermarx said...
I was expecting you'd say that but as I didn't know how to do it without starting over I hoped it was ok like this.

It is "ok", but the problem is that it is hard to update the changes that you have made against changes that have already been done in the develop branch.  For instance, @devlead has made a number of changes to the website, and as a result, this PR would need to be updated.  If the work was done on a "feature branch" it is really simple, however, when done on the develop branch, it is much, much harder.

PR #190, opened 2016-10-21
Add Cake.DoInDirectory

pitermarx:


gep13:
@pitermarx happy to pull this in, this time, however, I would strongly recommend that going forward, any changes you make are done in a branch off the develop branch.  Otherwise, it becomes very hard to update a PR that has become out of sync with the target branch.

Thanks for making this contribution!

PR #202, opened 2016-11-04
add Cake.Yarn AddIn

MilovanovM:
Support for [yarn](https://yarnpkg.com/) package manager

### DIFF
--- a/src/Cake.Web/App_Data/addins.xml
+++ b/src/Cake.Web/App_Data/addins.xml
@@ -605,4 +605,17 @@
     <Description>Cake AddIn which provides compression functionality for BZip2, GZip and Zip.</Description>
     <Categories>Compression</Categories>
   </Addin>
+  <Addin>
+    <Name>Cake.Yarn</Name>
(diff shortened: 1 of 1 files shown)


gep13:
@MilovanovM thanks for adding this!

This is now live on the site.  You can see it [here](http://cakebuild.net/addins/category/node) and [here](http://cakebuild.net/dsl/yarn).

One comment that I would make, is that I think the example documentation should be placed with each alias, rather than on the YarnRunner.

Right now, if you go here:

http://cakebuild.net/api/cake.yarn/932e1214/25a98705

You see all the examples.  But if you click through to here:

http://cakebuild.net/api/cake.yarn/0a112adc/29edf02b

Then there is no example.

Just a small comment though.  Thanks again for doing this!

MilovanovM:
@gep13 no problem, I've added the same code samples for YarnRunner methods. Does it make sense?

gep13:
@MilovanovM yes, that sounds good, thanks!

PR #209, opened 2016-11-11
Cake.SqlServerPackager

ilich:
Please add Cake.SqlServerPackager to the webite.


gep13:
Thank you for submitting this! 👍

gep13:
@ilich one comment...

Currently, the generated documentation looks like this:

image

If you include some additional information in your XML comments, you can make it look like this:

image

Notice the example script that shows how it can be used.  You can see the example here:

https://github.com/cake-contrib/Cake.Gem/blob/develop/Source/Cake.Gem/GemAliases.cs#L46

If you do decide to do this, pushing a new NuGet package is all that is required.  The additional example will be pulled through to the website the next time it is published.

Thanks!

PR #214, opened 2016-11-13
Added Cake.Watch to addins.xml

wk-j:
Added Cake.Watch


gep13:
@wk-j I have just rebased your branch, and I will merge once the CI builds complete.

One thing I noticed was this:

image

You are missing the `code` element within your `example` tag:

https://github.com/cake-addin/cake-watch/blob/master/Cake.Watch/WatchAlias.cs#L17

Have a look here as an example:

https://github.com/cake-contrib/Cake.Gem/blob/develop/Source/Cake.Gem/GemAliases.cs#L28

If you can fix that up, and push a new version of your nuget package, it will flow through onto the site.

Thanks for submitting this PR!

PR #222, opened 2016-11-19
Adding Cake.Stylecop to addins list.

Ashthos:
Ref: Published version 1.1.2 on nuget.


gep13:
@Ashthos thanks for getting this added!  Really appreciate it!

I raised one slight issue here: https://github.com/Ashthos/Cake.StyleCop/issues/7

If you fix that up before the next release of Cake.StyleCop, it will flow through automatically to the site, you won't have anything else to do, only to push a new NuGet package.

Issue #223, opened 2016-11-21
Add documentation for CakeNamespaceImportAttribute


kcamp:
As part of the work done for https://github.com/cake-build/cake/issues/1352 - Adding support for CakeNamespaceImportAttribute at the assembly level - documentation for the same was requested.

I have drafted the documentation and will submit PR shortly

phillipsj:
This can be closed.

gep13:
@phillipsj agreed

PR #224, opened 2016-11-21
GH223: Add documentation for the CakeNamespaceImportAttribute

kcamp:
This is a draft of the documentation for GH-223.

In the original thread at https://github.com/cake-build/cake/issues/1352 it was suggested that this documentation be added
>@kcamp @gep13 this would be under
>http://cakebuild.net/docs/contributing/documentation

It seems that this page relates more to documenting the code itself, and isn't as relevant to a user that might be developing Cake add-ins or script aliases.

For this PR, I added documentation to the page at http://cakebuild.net/docs/fundamentals/aliases
This seems more topically relevant and like it fits better, but I can move it anywhere that we feel it would be serve the community.

If there is feedback on the content, of if you'd like it relocated to the page as originally mentioned, let me know what changes are desired and I'll get them made.

Thanks!

### DIFF
--- a/input/docs/fundamentals/aliases.md
+++ b/input/docs/fundamentals/aliases.md
@@ -59,7 +59,32 @@ public static class MyCakeExtension
 }
 ```
 
-# Using the alias
+=======
+#### Importing Namespaces
+Your script alias may need to import one or more namespaces into your Cake script.  Cake supports the automatic import of namespaces with attributes.
+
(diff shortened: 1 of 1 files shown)


gep13:
@cake-build/team this looks good to me, and I think the choice of the page makes sense.

Thoughts?

If everyone is happy, when we release 0.18.0, we can pull in this documentation too.

phillipsj:
I think this can be closed.

PR #225, opened 2016-11-24
Add Cake.Sonar plugin information

tomstaijen:


tomstaijen:
image
Is my introduction text too long?

gep13:
@tomstaijen thanks for getting this submitted!

Personally, I think that there is never enough documentation, but I do agree that in this case it is quite a bit more than is seen when compared to the other addins.  I don't think this is a bad thing though.

@devlead @patriksvensson @agc93 what are your thoughts?

agc93:
Does the website render `<example>` tags in that summary view? Perhaps we could move the code blocks out of the `summary` and into `example`?

daveaglick:
The `SqlServer` addin summary is pretty long too. IMO, they're both a bit much for the adding/category pages that display more than one addin.

My recommendation is to move a chunk to the `<example>` element and then add rendering for that to the addin details view in vNext (I.e., show the example content here: http://cakebuild.net/dsl/sqlserver).

gep13:
@tomstaijen would you be able to update this PR with regard to the suggestions above?  And we can look at changing the website to ensure that the information flows through correctly?

@trailmax you might want to take a look at this as well.  Thanks!

tomstaijen:
sure thing

trailmax:
@gep13 Noted. I've seen the text spilling over and indeed there are too much info in some of the comments. I'll go update some docs just now

gep13:
>@trailmax said....
I'll go update some docs just now

Thank you!

**NOTE:** The changes won't be reflected on the live site until the next time we do a deployment to the live server.

trailmax:
@gep13 updated the nuget. Moved some of xml comments into Readme.md. New nuget package pushed for your build script to pick up -)

gep13:
@tomstaijen thanks for making the changes!

PR #229, opened 2016-11-29
Adding Cake.AppPackager AddIn

phillipsj:


gep13:
@phillipsj happy to pull this in.

One note for future PR's though... Recommendation is to create a feature branch off of develop to do the work on, rather than directly on your develop branch.  That way, if any changes are required they can be done more easily.

PR #237, opened 2016-12-21
Added Cake.Paket to addins list.

larzw:


larzw:
Added Cake.Paket to addins list...

Thanks!

devlead:
Hi @larzw this now merged, but I had to fix a typo in the addin assembly/xml filter so it'll be avail at cakebuild.net/dsl/paket shortly when I merge next PR.

PR #241, opened 2017-01-01
Fix GitHub profile link

agc93:
Changes name link to my GitHub profile link to fix 404


gep13:
@agc93 Ah, good catch! I think the URL that @devlead used only works for people who are in the organization.

gep13:
Although, that doesn't explain why it 404'ed for you. Unless you weren't logged in at the time.

agc93:
I was logged in, but I'm thinking it might be an owner vs member thing?

PR #242, opened 2017-01-01
Add vector logo

JimiC:
SVG version of logo, courtesy of me.


gep13:
@JimiC thank you for your interest, but I believe we already have this covered here: https://github.com/cake-build/graphics

PR #307, opened 2017-03-18
Proposed change: Add Cake.SynVer, remove Cake.SemVer.FromAssembly

wallymathieu:
Cake.SemVer.FromAssembly is now obsolete. The role has been taken over by Cake.SynVer.


gep13:
@wallymathieu thank you for keeping us up to date with this.

To avoid the duplication of the NuGet packages, can I suggest that you follow the deprecation process outlined here:

https://chocolatey.org/docs/how-to-deprecate-a-chocolatey-package

Yes, that is the process followed for Chocolatey Packages, but the same process applies equally to NuGet Packages.

PR #308, opened 2017-03-20
Adding Active Directory Addin

phillipsj:


gep13:
@phillipsj thank you for doing this!

PR #317, opened 2017-03-22
Add XComponent Cake AddIn

Julien-Molina:


gep13:
@Julien-Molina thank you for making this addition! 👍

PR #329, opened 2017-03-30
Spelling fix

daveaglick:


gep13:
@daveaglick thanks for fixing that.  Deploying to live site now.

PR #367, opened 2017-06-14
Script Aliases: fix header section formatting

patridge:
I wasn't totally sure on the intent of the original Markdown, but it looked like the heading syntax was being ignored or improperly rendered because of the prior line. Oddly, the existing `####` wasn't being rendered as plain text either.

Currently:

image

Removing the `=======` appears to give the H4 line it's formatting, though maybe the original intent was to have a horizontal line and H4 header there like this (e.g., `* * *`):

* * *
#### Importing Namespaces

### DIFF
--- a/input/docs/fundamentals/aliases.md
+++ b/input/docs/fundamentals/aliases.md
@@ -59,8 +59,8 @@ public static class MyCakeExtension
 }
 ```
 
-=======
 #### Importing Namespaces
+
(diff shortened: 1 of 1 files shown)


mholo65:
@patridge  this has been merged with this commit: 0ab065e

As a result, this PR will need to be manually closed, since GitHub doesn't follow the merge that was done. Don't worry, though, your work has been correctly attributed. Thanks!

PR #378, opened 2017-07-07
Include build.sh in getting-started.md

slang25:
Include both bootstrapping scripts so that people are encouraged to be cross platform by default.


gep13:
@slang25 thank you for taking the time to make this change!

Issue #386, opened 2017-07-18
[Documentation] Sample comment XML not rendering correctly on cakebuild.net


patridge:
Page: http://cakebuild.net/docs/contributing/documentation
Source: https://github.com/cake-build/website/blob/develop/input/docs/contributing/documentation.md

<img width="975" alt="screen shot 2017-07-18 at 15 46 56" src="https://user-images.githubusercontent.com/713665/28341316-5fa4a3ec-6bd0-11e7-8fb5-6f1e608d02d0.png">

Although GitHub is rendering it correctly, the XML code comments aren't always rendering correctly on the production page.

**Properties** and **Namespaces** seem fine, but **Classes** and **Methods** don't render the various XML comment tags.

## Random guesses

Who knows what sort of small thing might be setting it off. There are a few random inconsistencies that could compound to somehow break those XML comments, but it would be a pretty obscure bug if that's the case.

* [Extra space in link text: `[CakeAliasCategoryAttribute ](…)`](https://github.com/cake-build/website/blame/develop/input/docs/contributing/documentation.md#L10)
* [No line break between paragraph and start of code block](https://github.com/cake-build/website/blame/develop/input/docs/contributing/documentation.md#L10)
* [No line break between header and paragraph](https://github.com/cake-build/website/blame/develop/input/docs/contributing/documentation.md#L60)

daveaglick:
It's definitely odd that it's working for some blocks (Properties) but not other ones (Methods). First thing to look at is if it's a bug in Markdig or Wyam...

daveaglick:
I can confirm by feeding the raw MD file through Markdig in LINQPad that Wyam must be doing something after Markdown rendering to mess this up. I'll take a closer look tomorrow.

gep13:
@patridge the changes you made in #387 didn't seem to make any difference:

http://develop.cakebuild.net/docs/contributing/documentation

But thanks for trying 👍

daveaglick:
Pretty sure this is the same problem as reported in Wyamio/Wyam/issues/550 and is now fixed with Wyam 0.18.6. I'll circle back after the next regular publish to verify.

daveaglick:
Confirmed fixed.

PR #387, opened 2017-07-18
Normalize some Markdown whitespace.

patridge:
This is just a random shot-in-the-dark to try to fix comment XML rendering for #386. It's likely not going to fix the issue, but it shouldn't make it any worse, and makes the Markdown a little more consistent.


gep13:
@patridge thanks for taking the time to fix this.  As you say, I like the consistency this PR brings, even if it doesn't help with the underlying problem.

PR #396, opened 2017-08-22
Add an new Blogentry and correct moved URL

retschgi:
We moved our Domain from www.software-architekt.ch to bambit.ch. Therefore I corrected the Link and the Heading.
Additionally we've written another Blogpost about Cake. I added this Blogpost to the LinkList.


gep13:
@retschgi Vielen Dank!

PR #399, opened 2017-08-30
Need to use a double dash prefix for args instead of a single dash

blebaron:


gep13:
@blebaron does the sample, as is, not work for you?  I have just tested it out, and it seems to work fine.

blebaron:
@gep13 yesterday I couldn't get it working with the single dash and looking at [this post](https://stackoverflow.com/questions/40046752/how-do-i-pass-my-own-custom-arguments-to-build-ps1) from @devlead on SO, made me believe that the docs were wrong. Trying it again today it is working as documented. I'm guessing that I had left off the double quotes, not thinking that they were required where my arg didn't have any spaces. It is interesting that both of the examples below work.

Documented: `build.ps1 -ScriptArgs '-output="apps.zip"'`
Not Documented: `build.ps1 -ScriptArgs '--output=apps.zip'`

PR #418, opened 2017-10-10
[fundamentals] Add macOS example for setting arguments

patridge:
Since I was having trouble figuring out how to pass arguments to a Cake script on a Bash environment, I figured I'd add an example in the first place I looked unsuccessfully.


devlead:
@patridge your changes have been merged, thanks for your contribution 👍

PR #422, opened 2017-10-16
Fixed 'IsDependencyOf' blog typo

jnm2:
We never had `IsDependencyOf`, sadly, but if we did, it would by a synonym for `IsDependeeOf`. This example meant to use `IsDependentOn` as the antonym.


devlead:
@jnm2 your changes have been merged, thanks for your contribution 👍

PR #424, opened 2017-10-18
Limits DSL categories to distinct values (#356)

### LINKED ISSUE #356: Exclude duplicate classes from DSL category pages
It's currently possible to get duplicate class symbols for the same class when scanning for `CakeAliasCategoryAttribute`. For example, if multiple attributes with the same category are applied to a single class or applied to multiple partials of the same class. If that happens, we currently treat the duplicates as separate classes and end up generating duplicate method docs. See http://cakebuild.net/dsl/docker/.

To resolve, just need to make sure each class symbol is distinct for the DSL category page.

daveaglick:
Resolves #356. Even though there's no longer a problem in the Docker alias (that was resolved by removing the redundant `CakeAliasCategoryAttribute` directly), you can still see the problem in action at http://cakebuild.net/dsl/signing/

You don't pay me the big bucks for adding a single line, you pay me the big bucks for knowing which line to add.


devlead:
I'm taking this for a spin now.

devlead:

PR #435, opened 2017-11-09
Removed obsolete section from VSO docs

Cheesebaron:
The steps with pinning nuget.exe and changing the bootstrapper are no longer relevant.


gep13:
@Cheesebaron can you elaborate on why this is no longer required?

@agc93 thoughts?

Cheesebaron:
The latest NuGet that the default bootstrapper gets is currently 4.3.x, when that piece of doc was written in was some 3.x variant.

I have multiple VSTS builds running without these modifications.

gep13:
@Cheesebaron your changes have been merged, thanks for your contribution 👍

PR #444, opened 2017-12-01
Fix grammar

nafarlee:


gep13:
@nafarlee thank you for taking the time to fix this, it is very much appreciated!

PR #447, opened 2017-12-02
Adding yml for Cake.Talend

phillipsj:


gep13:
@phillipsj Thanks for doing this!  Is this the first of many? 😄

Issue #416, opened 2017-09-28
Document DoesForEach & DeferOnError


devlead:
cake-build/cake#1767 introduced DeferOnError and DoesForEach 

Should be documented on the website.

Samples here:
https://github.com/cake-build/cake/blob/develop/tests/integration/Cake.Core/CakeAliases.cake

```cake
Task("Awesome")
    .Does(() => { throw new Exception(); })
    .DoesForEach(GetFiles("**/*.txt"), (file) => { // My cool file action })
    .DeferOnError();
```

phillipsj:
@devlead I am going to take this one.

PR #454, opened 2017-12-09
Adding docs for DoesForEach and DeferOnError

phillipsj:
Let me know if their at changes. #416


gep13:
@phillipsj thanks for doing this!

PR #466, opened 2017-12-31
feat(criteria docs): add pipeline with criterias

jrgcubano:
Implements #467

### DIFF
--- a/input/docs/fundamentals/criteria.md
+++ b/input/docs/fundamentals/criteria.md
@@ -7,19 +7,126 @@ You can control and influence the flow of the build script execution by providin
 Task("A")
     .WithCriteria(DateTime.Now.Second % 2 == 0)
     .Does(() =>
-{
-});
-
+    {
+    });
+    
 Task("B")
     .WithCriteria(() => DateTime.Now.Second % 2 == 0)
     .IsDependentOn("A")
-    .Does(() =>
-{
-});
+    .Does(() => 
+    {
+    });
 
 RunTarget("B");
 ```
 
 Task `A`'s criteria will be set when the task is created while Task `B`'s criteria will be evaluated when the task is being executed.
 
+In a real project, we would control the execution of the tasks as a pipeline according to different environmental states, the repository branches, etc. In this case we get two criterias from outside to detect if local build and we are working on master branch. Notice we are using AppVeyor Build System, you can do similar stuff with other Build Systems as well.
+
+```csharp
+var isLocalBuild = BuildSystem.IsLocalBuild
+var isMasterBranch = StringComparer.OrdinalIgnoreCase.Equals("master",
+    BuildSystem.AppVeyor.Environment.Repository.Branch));
+    
+var deployUrl = HasArgument("DeployUrl") 
+    ? Argument<string>("DeployUrl") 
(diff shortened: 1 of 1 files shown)


input/docs/fundamentals/criteria.md
```diff
@@ -6,20 +6,113 @@ You can control and influence the flow of the build script execution by providin
 ```csharp
 Task("A")
     .WithCriteria(DateTime.Now.Second % 2 == 0)
-    .Does(() =>
-{
-});
-
+    .Does(() => ... );
```
gep13: I am not sure about the inclusion of the ... here.  Some might think that this is part of the Cake DSL.  I would prefer that this is left as it was.

input/docs/fundamentals/criteria.md
```diff
@@ -6,20 +6,113 @@ You can control and influence the flow of the build script execution by providin
...
+    
 Task("B")
     .WithCriteria(() => DateTime.Now.Second % 2 == 0)
     .IsDependentOn("A")
-    .Does(() =>
-{
-});
+    .Does(() => ... );
```
gep13: Same as the above comment.

input/docs/fundamentals/criteria.md
```diff
@@ -6,20 +6,113 @@ You can control and influence the flow of the build script execution by providin
...
+    BuildSystem.AppVeyor.Environment.Repository.Branch));
+    
+var deployUrl = HasArgument("DeployUrl") 
+    ? Argument<string>("DeployUrl") 
+    : EnvironmentVariable("DEPLOY_URL");
+    
+var target = "Default";
+
+Task("Clean")
+    .Does(() => ... );
```
gep13: I see what you are trying to do here, but I think overall it would be clearer if you included the execution of the lambda, even if it is a single `Information` method.

input/docs/fundamentals/criteria.md
```diff
@@ -6,20 +6,113 @@ You can control and influence the flow of the build script execution by providin
...
-{
-});
+    .Does(() => ... );
 
 RunTarget("B");
 ```
 
 Task `A`'s criteria will be set when the task is created while Task `B`'s criteria will be evaluated when the task is being executed.
 
+In a real project, we would control the execution of the tasks as a pipeline according to different environmental states, the repository branches, etc. In this case we get two criterias from outside to detect if local build and we are working on master branch.
```
gep13: You may want to call out the fact that you are specifically using AppVeyor, but you can do similar stuff with other Build Systems as well.

input/docs/fundamentals/criteria.md
```diff
@@ -6,20 +6,113 @@ You can control and influence the flow of the build script execution by providin
...
+```csharp
+...
+var target = "Default";
+
+var isRecoveryMode = false;    
+
+Task("Migrate")    
+    .WithCriteria(!isLocalBuild)
+    .WithCriteria(isMasterBranch)
+    .Does(() => ... )
```
gep13: Same as comments above.

jrgcubano:
@gep13 done with corrections!

gep13:
@jrgcubano thanks for doing this!

PR #479, opened 2018-01-16
Assembly name was incorrect

JonCubed:
@gep13 would this explain why it wasn't showing up on the website?


gep13:
@JonCubed yes, that would explain it.  The documentation for the site is extracted from the XML Comments of the assembly that is generated.  If this name isn't correct, then the documentation generation will silently move onto the next addin.

devlead:
@JonCubed your changes have been merged, thanks for your contribution 👍

PR #445, opened 2017-12-02
[WIP] Added FAQ section

gep13:
- Including new top level menu item

PR #516, opened 2018-05-11
Revised FAQ section

daveaglick:
This PR takes the place of #445 and includes the following additional changes:
* Rebased the PR to latest develop
* Added anchorjs library (note that this applies everywhere, looks pretty good IMO)
* Fixes the anchor links on the FAQ page


gep13:
@cake-build/cake-team any objections to getting this pulled in, and then we can start to expand on the content?

gep13:

gep13:
@cake-build/cake-team it is now live on the develop site:

https://develop.cakebuild.net/faq/

There is at least one broken link that will need to be corrected, but I think it is still a good start, and anytime we get a commonly asked question, we should add an entry onto this page, so that it isn't lost.

PR #531, opened 2018-06-18
Fix inconsistent casing and typos in YML files

Jericho:


gep13:
@Jericho thanks for fixing this!

PR #550, opened 2018-07-02
[TESTING, PLEASE IGNORE] Update YAML files

Jericho:
Resolves #549

PR #548, opened 2018-06-28
[TESTING, DO NOT MERGE] Update YAML files

Jericho:
Resolves #547

PR #555, opened 2018-08-20
Update Wyam to 1.5.0

daveaglick:
Verified with a full local build, everything seems to be working as expected. This should unblock @Jericho and #548 and #550.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,8 +1,8 @@
 #tool "nuget:https://api.nuget.org/v3/index.json?package=KuduSync.NET&version=1.3.1"
-#tool "nuget:https://api.nuget.org/v3/index.json?package=Wyam&version=1.2.0"
+#tool "nuget:https://api.nuget.org/v3/index.json?package=Wyam&version=1.5.0"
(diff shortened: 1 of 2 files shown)


build.cake
```diff
@@ -194,17 +194,17 @@ Task("Preview")
 Task("Debug")
     .Does(() =>
     {
-        StartProcess("../Wyam/src/clients/Wyam/bin/Debug/wyam.exe",
-            "-a \"../Wyam/src/**/bin/Debug/**/*.dll\" -r \"docs -i\" -t \"../Wyam/themes/Docs/Samson\" -p");
+        StartProcess("../Wyam/src/clients/Wyam/bin/Debug/net462/wyam.exe",
+            "-a \"../Wyam/tests/integration/Wyam.Examples.Tests/bin/Debug/net462/**/*.dll\" -r \"docs -i\" -t \"../Wyam/themes/Docs/Samson\" -p");
```
daveaglick: The paths here had to change since the client app now uses an SDK-style project file which puts binaries under a target framework subdirectory. Figured this was loosely related to bumping the version so I left it in here instead of adding another PR. I'm probably the only one who uses this build target anyway.

devlead:

PR #569, opened 2018-09-07
Mention --exclusive in the Running Targets page

jnm2:


input/docs/fundamentals/running-targets.md
```diff
@@ -34,8 +34,14 @@ Task("Publish")
...
+With this Cake script, you can run a specific target by passing the `-Target` argument to `Cake.exe`. Thus, we can run the `"Publish"` target by calling:
 
 ```powershell
 ./build.ps1 -Target Publish
 ```
+
+The `--exclusive` parameter causes `RunTarget` to run only the specified target and no dependencies. This command runs the Publish target without running the Build target:
+
+```powershell
+./build.ps1 -Target Publish -Exclusive
```
gep13: @jnm2 any reason that you chose to use `-Exclusive` here, rather than `--exclusive`

jnm2: To match the `-Target` argument on the same line. Should I change both to dash-dash-lower?

gep13: Currently, the `-Target` is covered by a parameter in the PowerShell bootstrapper.  There isn't the same thing for the `--exclusive` argument, so I think it would be more correct to use `--`.  We, the Cake Team, have spoken about being consistent everywhere and always using `--` though, so perhaps better to use it on both arguments.

jnm2: Oh, I see. I added it to my bootstrapper, but I forgot that cake.exe doesn't recognize the single-dash variants without bootstrapper interpretation.

gep13:
@jnm2 your changes have been merged, thanks for your contribution 👍

PR #579, opened 2018-10-02
Add note that .NET Framework 4.7.2 is suggested for running .NET Standard addins

pascalberger:
Add note that .NET Framework 4.7.2 is suggested for running .NET Standard addins


devlead:
@pascalberger  your changes have been merged, thanks for your contribution 👍

PR #588, opened 2018-10-31
Added missing addins.

agc93:
Just realised I hadn't added a couple of the more recent addins, so just small additions :)


devlead:
@agc93 your changes have been merged, thanks for your contribution 👍

PR #591, opened 2018-11-02
Update for Azure DevOps

pascalberger:
Replace branding from VSO / VSTS to Azure DevOps. Also update section about secret variables which now can be mapped using YAML build definitions.

### DIFF
--- a/input/docs/build-systems/azure-devops.md
+++ b/input/docs/build-systems/azure-devops.md
@@ -1,18 +1,18 @@
 Order: 10
-Title: Visual Studio Team Services
+Title: Azure DevOps
 ---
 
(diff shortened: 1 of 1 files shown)


gep13:
@devlead @pascalberger  Hmm, I "think" there was a reason why we hadn't updated from vso to vsts.  @patriksvensson do you remember?

On this page:

https://marketplace.visualstudio.com/items?itemName=cake-build.cake

We would need to update the link.  In the meantime, can we put a re-direct in place from vso to the new page?

devlead:
@gep13 I'm adding that.

devlead:
@gep13 I've added the redirect with 651a4a23, I'll merge to master if looks good on develop Azure.

PR #678, opened 2019-07-27
Update Cake.BitDiffer.yml

WebDucer:
Update repository url after moving the repository to cake-contrib organization

### DIFF
--- a/addins/Cake.BitDiffer.yml
+++ b/addins/Cake.BitDiffer.yml
@@ -2,7 +2,7 @@ Name: Cake.BitDiffer
 NuGet: Cake.BitDiffer
 Assemblies:
 - "/**/Cake.BitDiffer.dll"
-Repository: https://github.com/WebDucer/Cake.BitDiffer
+Repository: https://github.com/cake-contrib/Cake.BitDiffer
 Author: Eugen [WebDucer] Richter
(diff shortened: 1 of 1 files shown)


Jericho:
@WebDucer FYI: we have an automated process to maintain YAML files on Cake's website. This process uses your addin's nuget package metadata to generate the YAML content and therefore any manual modification to the YAML file can potentially be overwritten if it doesn't match what's in the package metadata. I say "potentially" because the automated process only update certain properties such as project URL and categories for example which means that, depending which property you manually modify, it may or may not get overwritten.

In your case there is a discrepancy between the project URL in your package metadata (see the `Project Site' link [here](https://www.nuget.org/packages/Cake.BitDiffer/)) and the value you just manually updated via this pull request.

There is, however, one exception (which applies here): our automated process is smart enough to recognize when a given addin has been moved to the cake-contrib organization and the project URL in the metadata hasn't been updated accordingly, which is exactly your situation. In this case, the URL under that cake-contrib is favored over the URL under the original author's personal github repo.

To make a long story short: as a rule of thumb, it's preferable to maintain the nuspec in your addin's nuget package and allow our automated process to update the YAML file as necessary.

WebDucer:
@Jericho The update of the package (with changes from the move to cake-contrib) is on the way.

PR #688, opened 2019-08-20
fix broken link

kjjuno:


gep13:
@kjjuno thank you for finding and fixing that, appreciate it!

PR #725, opened 2019-09-19
Conflicting statements w.r.t RunTarget and Setup

Jimit09:
Below statements described in the document is conflicting.

Setup will only be called if a call to `RunTarget` is made and the dependency graph is correct.
If `RunTarget` is called before the `Setup` or `Teardown` methods are called, they won't be correctly setup and won't work.

Setup will not be called until RunTarget is called and If `RunTarget` is called before the `Setup` or `Teardown` methods are called, they won't be correctly setup and won't work.

What's the correct dependency?

### DIFF
--- a/input/docs/fundamentals/setup-and-teardown.md
+++ b/input/docs/fundamentals/setup-and-teardown.md
@@ -38,4 +38,3 @@ TaskTeardown(teardownContext =>
 
 ```
 
-**NOTE:** If `RunTarget` is called before the `Setup` or `Teardown` methods are called, they won't be correctly setup and won't work.


gep13:
@cake-build/cake-team anyone know an answer to this one off the top of their head?

patriksvensson:
@Jimit09 @gep13 `RunTarget` must be called after `Setup` and `Teardown` as it is now. We've been talking about making RunTarget implicit, but we would need to spec an RFC for that in that case.

input/docs/fundamentals/setup-and-teardown.md
```diff
@@ -38,4 +38,3 @@ TaskTeardown(teardownContext =>
 
 ```
 
-**NOTE:** If `RunTarget` is called before the `Setup` or `Teardown` methods are called, they won't be correctly setup and won't work.
```
devlead, line 41: This statement is correct so thing it should remain.

gep13:
@Jimit09 we have discussed this as a team, and we feel that this information should remain, however, it may need to be reworked as part of a larger documentation effort.  This should hopefully be happening soon, so keep an eye out for those changes.

PR #730, opened 2019-10-01
Updates Wyam to 2.2.8 to mitigate ongoing package installation issues

daveaglick:
The new Azure Pipelines images with latest VS break Wyam package installation due to outdated NuGet client library incompatibilities. This updates Wyam to avoid those errors.

I also updated the readme to account to argument differences now that we're building with `Cake.Tool`.


devlead:

PR #770, opened 2019-10-14
Remove Cake.Tfs addin since it is deprecated

pascalberger:
Remove Cake.Tfs addin since it is deprecated


pascalberger:
@Jericho https://github.com/cake-contrib/Cake.AddinDiscoverer/issues/96 is already released (not that Addin Discoverer will add it again once removed)?

Jericho:
Not only will AddinDisco NOT add a YAML file for this addin but also it will also suggest to delete the YAML file during the next schedule build (which is midnight UTC) so there is no need to manually delete the YAML file.

pascalberger:
> Not only will AddinDisco NOT add a YAML file for this addin but also it will also suggest to delete the YAML file during the next schedule build (which is midnight UTC) so there is no need to manually delete the YAML file.

Oh, nice. That would have been my next suggestion otherwise :)

@Jericho You sure it is working as expected? Because I marked it as deprecated over 24 hours ago. 

If you want to use this as a test case I can also delete the PR and let addin discoverer do its magic.

Jericho:
I know it has worked in the past with other deprecated addins so I'm surprised that it didn't work in this case. Hold on, let me investigate real quick...

Jericho:
It works: https://github.com/cake-build/website/issues/771

I wonder if there was a delay for the "deprecated" flag to be propagated on nuget.org???

gep13:
LGTM!

PR #775, opened 2019-10-15
Spelling mistake corrected

SupriyoSarkar:
In source code folders spelling mistake is corrected from Addins to Addons


gep13:
@SupriyoSarkar thank you for taking the time to raise this PR, however, this isn't a valid change.  We refer to these as addins throughout the Cake documentation and source code.  It you can point me to somewhere that this isn't the case, we can definitely get this updated.

In future, can I suggest that you raise an issue about a change that you would like to make so that we can discuss changes? That way, their won't be any wasted effort on your part.  Thanks.

SupriyoSarkar:
Yeah, sure . I appreciate that .

PR #844, opened 2020-01-26
Adding yml for new addin - Cake.SonarResults

rhtnr:
Adding yml for new addin - Cake.SonarResults 
Addin can break builds if SonarQube analysis fails after running a SonarQube analysis

### DIFF
--- a/addins/Cake.SonarResults.yml
+++ b/addins/Cake.SonarResults.yml
@@ -1,0 +1,9 @@
+Name: Cake.SonarResults
+NuGet: Cake.SonarResults
+Assemblies:
+- "/**/Cake.SonarResults.dll"
+Repository: https://github.com/rhtnr/Cake.SonarResults
+Author: Rohit Nair
+Description: Cake Addin to break the build after running the sonnar-scanner tool.
+Categories:
+- code quality


rhtnr:
@Jericho 
Thanks for clearing that up.
I had originally navigated to the instructions from https://cakebuild.net/addins/
I had understood that the a new YAML file had to be created. I didn't find anything on the automated process in the link.

Jericho:
Thanks for pointing me to the outdated content. I submitted a [PR](https://github.com/cake-build/website/pull/845) to clarify the instructions

Jericho:
One more thing I meant to mention: there are already two addins ([here](https://www.nuget.org/packages/Cake.Sonar/) and [here](https://www.nuget.org/packages/Cake.SonarScanner/)) that seem to provide same or similar functionality as the one you just created. What is different about the one you created?

pascalberger:
The other ones are calling different scanners (which are used to provide results to SonarQube), while this calls SonarQube to check for the calculated results. They do all different things, but for the same product. Therefore the question is if we should try to consolidate into one addin co-maintained by the different authors on the cake-contrib organization.

rhtnr:
As far as I understand, the addins listed above only initiates the analysis on the sonar server. My tool is a complementary addin to the addins listed above. The addin waits for the analysis to complete on the server, retrieves the results and sees if the analysis is a PASS or a FAIL depending on which the addin chooses whether to continue with the Tasks or break the build

Jericho:
thanks for clarifying. 

@pascalberger is right: we should consider merging the addins into a single one since they provide complimentary functionality.

Jericho:
Closing this PR because the automated tool has [discovered the new addin](https://github.com/cake-build/website/issues/846) and automatically generated the yaml file.

PR #914, opened 2020-06-21
Switch to registration5

mholo65:
This is a quick fix.

Correct fix would be instead of hardcoding, we should just read the `RegistrationsBaseUrl` from https://api.nuget.org/v3/index.json


gep13:
@mholo65 thanks for fixing this!

PR #926, opened 2020-06-27
Small improvements on wording on homepage

pascalberger:
Some small improvements on wording on homepage:

- Put .NET Core first and make clear that .NET means .NET Framework, compared to .NET 5
- Get rid of TFS, since it is now also called Azure DevOps
- Mention addin eco-system


gep13:
@pascalberger thanks for doing this!

PR #985, opened 2020-09-30
improved docs 

aaditya0909:


devlead:
Not clear what value this PR adds, going of to close it as invalid.

PR #994, opened 2020-09-30
Update readme footer

ShashankPawsekar:
Use the proper indication of year that ensure you are updated.


pascalberger (commented):
Sorry, but I don't see how this contributes to the documentation.

PR #1033, opened 2020-10-01
Make clear Frosting also works for .NET Framework

pascalberger:
Make clear Frosting also works for .NET Framework


gep13 (approved):
LGTM!

Issue #1044, opened 2020-10-03
Add documentation for Docker images


pascalberger:
We should add documentation for [Docker images](https://github.com/cake-build/docker)

NamanPatidar:
Yes i am interested .

Abhishekchauhan089:
I would like to work on this project

pascalberger:
For anybody interested in this: Please read first through [Cake Getting started section](https://cakebuild.net/docs/getting-started/running-cake-scripts) and [README](https://github.com/cake-build/docker) for Cake Docker images.

Scope of this issues is to document Docker images as a first class citizen for running Cake scripts, explaining the use cases for running Cake in a Docker container (mainly when using as a builder image). Having the possibility to run Cake in a Docker container should be documented at [Running Cake Scripts](https://cakebuild.net/docs/getting-started/running-cake-scripts) page. Since the Docker image is basically a deployment unit for an existing runner, we should add additionally somewhere a dedicated page explaining how to use the Docker image as a builder image for building Docker containers.

@NamanPatidar @Abhishekchauhan089  Please let me know if you're still interessted in working on this issue considering the above explained scope.

PR #1045, opened 2020-10-03
Add files via upload

NamanPatidar:
I resolved the issue regarding the "Add documentation for Docker images".


pascalberger:
@NamanPatidar This is not the kind of documentation we require. This is a website and we don't require a Word document. Also the document is for a random Docker container. We need documentation how to work with our specific Docker images. If you're still interested I can give you some hints  in #1044

Issue #179, opened 2016-10-13
Add compelling examples for end users.


reicheltp:
It would be awesome if the website would contain some complete and compelling examples for users who are new to Cake.

I have something in my mind like: 
- Build an Android APK and upload to HockeyApp
- Build a .NET Project with Semantic Versioning and upload artifacts to NuGET
- Run a bunch of Unit Tests and Upload Test Coverage to coverall.io
- ...

The samples could may be loaded from a community maintained repo in the cake-contrib organization. This increase the chance that the examples are always up to date.

pascalberger:
List of examples showing specific aspects:

- [ ] Preprocessor directives (https://github.com/cake-build/website/pull/1065)

pascalberger:
We also have an example repository: https://github.com/cake-build/example

PR #1075, opened 2020-10-07
Move documentation for runners in "Running Builds" section

pascalberger:
Move documentation for runners in "Running Builds" section. Previously we had a page listing the different runners and one explaining bootstrapping scripts. With this PR we have one page for every runner, containing requirements and bootstrapping documentation. 

image

We still have an overview about the capabilities of the different runners at `/docs/running-builds/runners`:

image

In subsequent PRs I would like to cleanup documentation further by moving documentation for CLI, using pre-release versions and pinning cake versions into documentation for every runner. This will allow to have these documentation available for every runner (currently it is only for .NET Framework and .NET Core runner, but not .NET Core tool or Frosting)


input/docs/running-builds/runners/dotnet-core-tool.md
```diff
@@ -0,0 +1,53 @@
...
+:::
+
+## Setup
+
+There's a one-time setup required for configuring a repository to use Cake .NET Core tool.
+
+Make sure to have a tool manifest available in your repository or create one using the following command:
+
+```shell
+dotnet new tool-manifest
```
gep13, line 27: @pascalberger not for this PR, but should we be explicit here and mention that the generated manifest file should be checked into source control, so that it can be used in subsequent builds?

gep13 (approved):
LGTM!

input/docs/running-builds/runners/dotnet-core-tool.md
pascalberger: 👍 I would like to add more tutorials which can show different aspects in depth, instead of just a single getting started. This would allow to document these kind of stuff. Added this point to https://github.com/cake-build/website/issues/179

gep13: Sounds like a plan.

PR #1077, opened 2020-10-07
Use Wyam compatible heading levels

pascalberger:
Start with H1 which will be shown in the right-hand navigation

### DIFF
--- a/input/docs/writing-builds/sharing-build-state.md
+++ b/input/docs/writing-builds/sharing-build-state.md
@@ -4,7 +4,7 @@ RedirectFrom: docs/fundamentals/sharing-build-state
 
 Sharing variables and state across your builds can currently be achieved in a handful of different ways, depending on what state you want to share and across what scope.
 
-## Typed Context
+# Typed Context
 
 :::{.alert .alert-success}
 Available since Cake 0.28.0.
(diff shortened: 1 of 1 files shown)


gep13:
@pascalberger I have often wondered about _what_ should be the first heading with markdown files, when there is a frontmatter title.  When you have a frontmatter title and you start with `#` markdownlint generates an error saying that this is incorrect.  In the case of this particular file, it doesn't have a Title in the frontmatter, so there won't be a warning raised, but I am curious about what is the _right_ thing to do here.

pascalberger:
>  but I am curious about what is the _right_ thing to do here.

From a semantic structure point of view documents should only have one H1 (which is the reason markdownlint complains), but Wyam looks for H1 to build the TOC on the right hand side. So while from a semantic point of view wrong, H1 is the correct thing to do in this cause because of tooling.

I don't know if there is a config setting for Wyam to check for H2 instead, but this would require changing all other pages, and IMHO out of scope for this PR.

@daveaglick Is there a config setting to have Wyam look for H2 instead of H1 to build the TOC?

gep13:
@pascalberger you can modify Wyam to include another level of headings in the right hand side using:

```
Pipelines["RenderPages"].InsertAfter("WriteMetadata", new Headings(3); // Heading level
```

pascalberger:
@gep13 OK, then we can either merge this and change afterwards to H2 on all pages. Or skip this and change to H2 on all pages directly in a separate PR.

gep13:
For a recent change that I did on another Wyam site, the above resulted in the following:

image

Where all headings `#` and `##` were then shown in the right hand side.

gep13 (approved):
LGTM!

gep13:
>@pascalberger said...
OK, then we can either merge this and change afterwards to H2 on all pages

Sounds like a plan to me.

pascalberger:
Follow up issue: https://github.com/cake-build/website/issues/1080

PR #1081, opened 2020-10-08
Use success annotations for highlighting positive aspects

pascalberger:
Use success annotations for highlighting positive aspects


gep13 (approved):
LGTM!

Issue #907, opened 2020-06-18
Document environment variable for showing what a tool is executing (Settings_ShowProcessCommandLine)


gep13:
This was added in this PR:

https://github.com/cake-build/cake/pull/2486

And I think we need to capture this, somewhere, on the site.

mgnslndh:
I can do this. Not sure where the info needs to be captured though. Perhaps on the page for [default configuration values](https://cakebuild.net/docs/fundamentals/default-configuration-values)? But this configuration is only available through an environment variable and not through the config file or command line I guess.

pascalberger:
Best solution would be to restructure content a little bit first (current structure shows its age 😄). My idea would be to have a `Configuration` section under `Documentation` (same level as `Fundamentals`) where the existing config values can be moved to, but also additional and currently missing configuration options be added. Once this is done, this can be added as a second page documenting environment variables. 

If this is OK for you I would suggest to do it in two pull requests: The first changing structure of the existing documentation and the second adding the content for this issue.

mgnslndh:
Sure, two PR is fine. 

# 1st PR

So we leave `fundamentals/configuration.md` as-is, explaining **how** configuration is done but we move the description of available configuration options to its own top level section. Or do you prefer having `fundamentals/configuration.md`  moved to something like `configuration/index.md`?

`fundamentals/default-configuration-values.md` is moved to `configuration/index.md` or `configuration/options.md` depending or you preference regarding my question above.

Missing configuration options needs to be identified and documented as well.

# 2nd PR

However, regarding the second PR splitting `fundamentals/default-configuration-values.md` into multiple pages i'm not sure. We would need to have three pages (file, env, cli) and each configuration option would need to have the same description and default value on each page. See mockups below.

### configuration/environment-variables.md
![image](https://user-images.githubusercontent.com/1139118/95045357-28b10f00-06e2-11eb-864c-a95bc428df6d.png)

### configuration/configuration-file.md
![image](https://user-images.githubusercontent.com/1139118/95045797-11beec80-06e3-11eb-9e14-4cef349cd3e6.png)

### configuration/command-line.md
![image](https://user-images.githubusercontent.com/1139118/95045938-5ea2c300-06e3-11eb-8797-b03eadd2b7fc.png)

pascalberger:
Sorry for the later reply. Missed your answer.

> 1st PR

[...]

PR #1087, opened 2020-10-08
Restructure configuration pages

mgnslndh:
@pascalberger:

> I would suggest to have docs/configuration/ giving a brief introduction to configuration for Cake, plus the list of sub-pages (like e.g. done at https://cakebuild.net/docs/extending/modules/). This would be in the index.cshtml file. I think the most part of docs/fundamentals/configuration.md can be moved to something like docs/configuration/set-configuration-values.md.

Part of #907

### DIFF
--- a/input/docs/running-builds/configuration/index.cshtml
+++ b/input/docs/running-builds/configuration/index.cshtml
@@ -1,0 +1,8 @@
+Order: 40
+---
+
+<p>
+    Cake can be configured so that you can control the internals of how Cake operates during the build.
+</p>
+
+@Html.Partial("_ChildPages")
--- a/input/docs/running-builds/configuration/default-configuration-values.md
+++ b/input/docs/running-builds/configuration/default-configuration-values.md
@@ -1,10 +1,12 @@
-Order: 50
-RedirectFrom: docs/fundamentals/default-configuration-values
+Order: 20
+RedirectFrom:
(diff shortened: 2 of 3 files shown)


input/docs/running-builds/configuration/index.cshtml
```diff
@@ -0,0 +1,9 @@
+Order: 40
+RedirectFrom: docs/running-builds/configuration
```
pascalberger, line 2: This redirects to itself and leads to an infinite redirect

pascalberger (changes requested):
There something wrong with one of the redirects. Otherwise LGTM

gep13:
Shouldn't the existing redirects remain in place? So that both old locations for the pages continue to get to the right place?

pascalberger:
> Shouldn't the existing redirects remain in place? So that both old locations for the pages continue to get to the right place?

Yes missed that one. We need to keep existing redirects and add additional redirects for the move in this PR

input/docs/running-builds/configuration/index.cshtml
mgnslndh: Yes, that does not seem right. I'll fix it.

pascalberger (approved):
LGTM

pascalberger:
@mgnslndh your changes have been merged, thanks for your contribution 👍

PR #1088, opened 2020-10-08
(GH-1085) Update CLI usage

### LINKED ISSUE #1085: Update usage information for different runner
https://cakebuild.net/docs/running-builds/cli/usage needs to be updated for the different runners / frosting. Also switches are outdated for current version

pascalberger:
Update CLI usage documentation. Prevoiusly documentation was for `Cake.exe`. Instead of a separate page usage instructions are now part of the documentation page of every runner. We can not use a separate page for usage instructions since Frosting probably always will have slightly different command line switches.

Switches for alle runners have been updated to the current released version. It also adds documentation for Cake Frosting. For easier maintaining of switches they have been moved into an external file.

Documentation how to read custom switches has been moved to http://localhost:5080/docs/writing-builds/args-and-environment-vars.

To keep documentation simple I removed examples for caling the runners.

Out of scope for this PR is to update documentation for version 1.0 and introduce a column to the switches for adding the version in which a switch was introduced (https://github.com/cake-build/website/issues/1089) and update of the environment variable part of http://localhost:5080/docs/writing-builds/args-and-environment-vars (https://github.com/cake-build/website/issues/1090).

image

Fixes #1085


patriksvensson (approved):

gep13:
@pascalberger @patriksvensson I didn't think that `--target` was an official argument, but rather one that is used by the script.

pascalberger:
@gep13 

Yes, this is what documentation says :)

> This is a custom switch which script must support explicitly.

This is also not something which has been added in this PR, but was there before.

gep13:
>@pascalberger said...
Yes, this is what documentation says :)

Can we make that clearer?  I missed that, and I was looking for it 😄

pascalberger:
> This is also not something which has been added in this PR, but was there before.

Sorry, thats not really true. What I did was to update switches with what `cake.exe --help` outputs. But I just saw that it has been removed for 1.0, so I would suggest to remove it also here.

pascalberger:
> Can we make that clearer? I missed that, and I was looking for it 😄

@gep13 See Make it clear that --target is not a switch of the runner

PR #1093, opened 2020-10-08
(GH-1071) Add sitemap generation pipeline

### LINKED ISSUE #1071: Add sitemap.xml
To improve indexing with search engines, and also for https://github.com/cake-build/website/issues/153, it would be helpful to have a sitemap.xml. Wyam has a [sitemap module](https://wyam.io/modules/sitemap).

gep13:
Based on example from here:

https://github.com/Wyamio/Wyam/blob/master/examples/Sitemap/config.wyam#L43-L51

Fixes #1071

### DIFF
--- a/config.wyam
+++ b/config.wyam
@@ -62,3 +62,9 @@ Pipelines.Add("RenderDsl",
     HtmlInsert("div#infobar-headings", (doc, ctx) => ctx.GenerateInfobarHeadings(doc)),
     WriteFiles()
 );
+
+Pipelines.Add("BuildSiteMap",
+    Documents(@ctx.Documents.ExceptPipeline("Addins").Except(@ctx.Documents.FromPipeline("Code"))),
+    Sitemap(),
+    WriteFiles((d,c) => "sitemap.xml")
+);


gep13:
@pascalberger looking at the generated sitemap.xml, it contains the yml files used to look for the addins, so there might be some more work that needs to be done here, but might need @daveaglick to provide some input here.

pascalberger:
@gep13 You can use `DocumentsFromPipelines(params string[] pipelines)` and pass in all pipelines whose documentes should be part of the sitemap. But IMHO this is not a very stable appraoch. What we instead want is all documents except the documents from the `Addins` pipeline. There's a `Documents().Where()` method and an `IDocumentColletion.ExceptPipeline` method, but I don't see yet how to make use of them. 

@daveaglick Any idea how to best filter out documents from a specific pipeline?

gep13:
@pascalberger adding `ExceptPipeline` into the mix seems to have the intended result.  No `yml` links are now included in the generated sitemap.xml file.

pascalberger (approved):
LGTM. We currently create some pages which we should not and which will be reachable through search engine once this is merged. I created https://github.com/cake-build/website/issues/1097 for this.

pascalberger (changes requested):
Just saw that there are other entries which link to non-existing pages:

- `https://cakebuild.net/SolutionInfo.cs` (all entries between `https://cakebuild.net/addins/*` and `https://cakebuild.net/api/*`)

`ExceptPipeline` doesn't accept multiple pipelines and can't be chained since it returns an `IEnumerable<IDocument>`. It should be possible though to filter further items with LINQ (similar to the [implementation of ExceptPipeline](https://github.com/Wyamio/Wyam/blob/develop/src/core/Wyam.Core/Documents/DocumentCollection.cs#L54))

gep13:
@pascalberger Do you know which pipelines are generating those files in order to exclude them, or were you think about excluding them based on URL?

pascalberger:
@gep13 Don't know for sure, but based on order, it might be the `code` pipeline.

gep13:
@pascalberger took another stab at this.  Can you let me know what you think about it?  Thanks

pascalberger (approved):
LGTM

pascalberger:
@gep13 your changes have been merged, thanks for your contribution 👍

PR #1094, opened 2020-10-08
Make it clear that --target is not a switch of the runner

pascalberger:
Make it clear that `--target` is not a switch of the runner, but something scripts often implement.


gep13 (approved):
LGTM!

PR #1102, opened 2020-10-09
(GH-1097) Use .txt extension rather than .md

### LINKED ISSUE #1097: Don't create pages for assets and shared folders
Currently pages are created for the `assets` and the `shared` folders:

- https://cakebuild.net/assets/
- https://cakebuild.net/shared/

While they are not reachable through navigation, they are part of sitemap.xml (https://github.com/cake-build/website/pull/1093) and therefore indexed for search endgines.

gep13:
Wyam, by default, will create pages for all .md files it finds.

There are a couple .md files that we don't want pages for.  Rather than
attempt to exclude them, handle them differently by changing the file
extension.

The referenced file is still added in as expected:

image

Fixes #1097


pascalberger (approved):
LGTM

gep13:
@pascalberger thanks for rebasing, I was just away to do this 😄

PR #1104, opened 2020-10-09
Update Cake.Unity3D.yml

mgnslndh:
Added missing package ID for Cake.Unity3D.yml

### DIFF
--- a/addins/Cake.Unity3D.yml
+++ b/addins/Cake.Unity3D.yml
@@ -1,5 +1,5 @@
 Name: Cake.Unity3D
-NuGet: 
+NuGet: Cake.Unity3D
 Assemblies:
 - "/**/Cake.Unity3D.dll"
 Repository: https://github.com/samoatesgames/Cake.Unity3D/


pascalberger (commented):
@mgnslndh Thanks for the pull request 👍  Addin yml files are updated automatically by a tool (https://github.com/cake-contrib/Cake.AddinDiscoverer/), so even if we merge this, it probably will be overwritten by the tool again.

@Jericho Any idea why the NuGet entry is missing here?

mgnslndh:
Ok. Another strange thing is the MagicChunks package which has a addin file named Magic-Chunks.yml

Other addins use the package ID in the file name.

mgnslndh:
According to this code it would seem that the name and NuGet package ID should be the same. Perhaps some other logic prevents this package to be synchronized but when it is, the YAML will be correct.

https://github.com/cake-contrib/Cake.AddinDiscoverer/blob/7ae4df91e0b0937ee9f94d10041074a07abfda4d/Source/Cake.AddinDiscoverer/Steps/SynchronizeYamlStep.cs#L228-L238

Jericho:
Regarding the `NuGet` package id: the AddinDiscoverer does not update this value in existing YAML files. We do set a value when generating a new YAML file, but we don't update the value in existing files. 

We cherry picked a few attributes in the YAML file to be automatically updated when we designed the discoverer and decided to leave any other value untouched because the files that existed at the time had been created by addin authors and we wanted to respect the values they had put in their YAML files (other than the ones we specifically decided to update of course).

The attributes that we update are:
- Assemblies
- Repository
- Prerelease
- Tags

The attributes that we do not update are:
- Name
- NuGet
- Author
- Description (we normalize line endings but otherwise we don't modify the content)

Having said that, you bring up an interesting scenario: an addin author set an attribute to a empty string in when creating their YAML file. Should the discoverer continue respecting this empty value or should we update the attribute?

`TL;DR;` : we don't update the NuGet attribute so changing it manually is the workaround.

pascalberger (approved):

pascalberger:
@mgnslndh your changes have been merged, thanks for your contribution 👍

PR #1106, opened 2020-10-10
Link to GitHub organization

pascalberger:
The "official" Cake ecosystem consists of more than Cake itself. Also integrations for editors and build servers, etc. We should therefore link to the GitHub organization which lists all of them instead of the Cake repo itself.


gep13 (approved):
LGTM

Issue #1090, opened 2020-10-08
Update documentation how to read environment variables


pascalberger:
[Documentation how to read environment variables](https://cakebuild.net/docs/writing-builds/args-and-environment-vars#environment-variables) in scripts needs to be updated for the different kind of runners.

PR #1113, opened 2020-10-11
(GH-1090) Update documentation how to read environment variables

pascalberger:
Updates documentation how to read environment variables:

image

I shortened content quite a bit, since IMHO stuff like how to set environment variables in different environments should be out of scope for our documentation (and most of the time build script need to read environment variables set by something else, e.g. build servers).

Fixes GH-1090


gep13 (approved):
LGTM

PR #1141, opened 2020-10-15
Use a boolean env variable to enable cache purging

pascalberger:
Use a boolean env variable to enable cache purging instead of just the existance of an env variable.

This allow to have a variable defined in the Azure Pipeline which can be set on a release.


devlead (approved):
LGTM 👍

devlead:
@pascalberger your changes have been merged, thanks for your contribution 👍

Issue #928, opened 2020-06-27
Add community / help section


pascalberger:
We should add a section about how to get help for users, linking to GitHub discussions, etc

karkranikhil:
Assign it to me i will try

pascalberger:
@karkranikhil Thanks for giving this a try 👍 

Currently we promote exclusively our [Gitter channel](https://gitter.im/cake-build/cake) on the website. We would like to move to [GitHub Discussion](https://github.com/cake-build/cake/discussions) going forward. Like the Gitter channel today, the idea of GitHub Discussion is to be a place for the community to connect with other community members about Cake.

For us it is important that we don't raise expectations that there is support on an enterprise level available, but it's a community effort. The new page should reflect this. Therefore I would suggest to introduce a new "Community" entry in the top level navigation, where in the future we can also bundle stuff like [Resources](https://cakebuild.net/docs/resources/), [Contribution info](https://cakebuild.net/docs/contributing/) (but this is out of scope of this issue).

Rough steps for this issues are:

* Remove the Open Chat window on the website
* Introduce a new community top level navigation entry
* Add a dedicated "Help" page to the community entry mentioning Discussions and Issues
* Mention Gitter as an archive on the "Help" page

pascalberger:
@karkranikhil Are you still interested in giving this a try? If you need any further information?

karkranikhil:
Sorry for the late reply. Can you give me the design?

pascalberger:
@karkranikhil What exactly do you mean with design? This shouldn't require any additional graphical design in the first step, just a new markdown page styled in the default styling of the website (see steps [here](https://github.com/cake-build/website/issues/928#issuecomment-702558183)).

PR #1145, opened 2020-10-17
Move FAQ page to documentation section

pascalberger:
Move FAQ page to documentation section to get more space in main navigation for other entries, like e.g. "Community" section (https://github.com/cake-build/website/issues/928).


pascalberger:
Don't know if it makes sense to keep the FAQ page with the current questions though. The answer to the pinning questions is just a link to the corresponding documentation at https://cakebuild.net/docs/writing-builds/reproducible-builds (the first sentence of the answer is no longer correct with support for the different runners and I would like to avoid duplicated docs).

The content of the last question is something I would like to add to https://cakebuild.net/docs/running-builds/.

Additionally we have search capability in the meantime which should make it easier to find the content.

gep13:
I have no objection to moving this to within the main docs, rather than the top level navigation.

I still think that an FAQ section is a good idea.  I see it more as an index of sort.  The ability to search the docs is great, but if you aren't sure what you are looking for, having the FAQs to fall back on is a great resource.  Most likely, the FAQ would just point to the relevant piece of the documentation.

Have a look at the Chocolatey FAQ's as an example:

https://chocolatey.org/docs/chocolatey-faqs

gep13 (approved):
LGTM!

gep13:
@pascalberger your changes have been merged, thanks for your contribution 👍

Issue #1096, opened 2020-10-09
Redesign / Mockup Addin section


pascalberger:
There are several issues regarding the addin section of the page:

- https://github.com/cake-build/website/issues/1095
- https://github.com/cake-build/website/issues/1078
- https://github.com/cake-build/website/issues/927
- https://github.com/cake-build/website/issues/686
- https://github.com/cake-build/website/issues/580
- https://github.com/cake-build/website/issues/299

Before starting to work on them it would make sense to start with a quick UI mockup considering the different open issues, so that we have a target to work against in the different issues.

pascalberger:
Suggestion for improvements to the current addin page. 

Menu item in the main navigation would be replaced from `Addins` to `Extensions` to also cover modules and recipe scripts in the future. 
The page will list all addins, modules and recipe scripts. For each entry it will show version (https://github.com/cake-build/website/issues/686), Cake compatibility (https://github.com/cake-build/website/issues/299), link to nuget.org and GitHub repository.
This page should be excluded from indexing for site-wide search engine.

![image](https://user-images.githubusercontent.com/2190718/96355711-e7a6fa80-10e4-11eb-9a4b-7d426ac4bcdd.png)

There's a search field which will perform a client side search on the listed extensions:

![image](https://user-images.githubusercontent.com/2190718/96355743-8c293c80-10e5-11eb-8135-5a2067aca85c.png)

There's also a filter button which toggles some filtering options. It allows to filter by different extension types, and a list of topcis (the current categories)

![image](https://user-images.githubusercontent.com/2190718/96355746-9f3c0c80-10e5-11eb-9775-b4a3857a69ac.png)

If user clicks on title of an entry in the list the detail page for an extension is shown.
It shows information how to use the extension with the different possibilities (https://github.com/cake-build/website/issues/1095) 
If it is an entry for an addin it will also show the aliases provided by the addin (https://github.com/cake-build/website/issues/927) (note that it shows the syntax to call it as an alias without the `ICakeContext` parameter).
These pages are indexed for the site-wide search engine.

![image](https://user-images.githubusercontent.com/2190718/96355764-bbd84480-10e5-11eb-908b-284925bc582c.png)

pascalberger:
Closing since all issues mentioned in the first post are now implemented

PR #1147, opened 2020-10-19
Move /addins to /extensions

pascalberger:
First (small) step on the way to design sketched at #1096. Current addin page should be able to host other extension (models, recipe scripts) and should therefore be available at `/extensions` instead of `/addins`.


gep13 (approved):
LGTM

gep13:
@pascalberger thanks

pascalberger:
PR to update search index: https://github.com/algolia/docsearch-configs/pull/2587

PR #1148, opened 2020-10-20
Add note that x.y.z needs to be replaced with actual Cake version

pascalberger:
Add note that x.y.z needs to be replaced with an actual Cake version


gep13 (approved):
LGTM

gep13:
@pascalberger thanks for getting this added!

PR #1157, opened 2020-10-24
Make sure column for alias list is wrapped

pascalberger:
Make sure column for alias list is wrapped when it contains a very long word without spaces.

Before:

image

After:

image


gep13 (approved):
LGTM

gep13:
@pascalberger thanks for fixing this!

PR #1158, opened 2020-10-25
Show label column only if there are labels set for an addin

pascalberger:
Show label column only if there are labels set for an addin.

Before:

image

After:

image


gep13 (approved):
LGTM!

gep13:
@pascalberger would be happy to see this merged after a rebase.

PR #1159, opened 2020-10-25
Remove Prerelease flag from addin metadata

pascalberger:
Remove Prerelease flag from addin metadata since it currently is not used. If we want to support it in the future again we should design implementation from scratch. Currently it just creates unecessary noise with PRs created by addin discoverer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,6 @@ The format of an addin file generally looks like:
 ```
 Name: Cake.Wyam
 NuGet: Cake.Wyam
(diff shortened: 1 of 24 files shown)


pascalberger:
Before merging addin discoverer needs to be updated to no longer set prerelease flag

Jericho (approved):
I am merging this PR and will publish a new version of the discoverer that no longer maintains the `prerelease` flag in the YAML files

PR #1209, opened 2020-11-03
Add redirects for every OS

pascalberger:
Add additional redirects for every OS & runner

### DIFF
--- a/download/bootstrapper/dotnet-core/linux/index.aspx
+++ b/download/bootstrapper/dotnet-core/linux/index.aspx
@@ -1,0 +1,1 @@
+<% Response.Redirect("https://raw.githubusercontent.com/cake-build/resources/master/dotnet-core/build.sh") %>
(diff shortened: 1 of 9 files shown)


gep13:
@pascalberger why are these necessary?

If I recall correctly, the original osx/windows/linux URL's were used for a while, but they were then replaced with the powershell/bash ones.  Those original redirects were left in place in case anyone was using them, but everywhere that we "owned" starting using the powershell/bash links.

pascalberger:
> @gep13 said...
> If I recall correctly, the original osx/windows/linux URL's were used for a while, but they were then replaced with the powershell/bash ones. Those original redirects were left in place in case anyone was using them, but everywhere that we "owned" starting using the powershell/bash links.

Currently in docs the OS ones are used for .NET Framework (https://cakebuild.net/docs/running-builds/runners/cake-runner-for-dotnet-framework#bootstrapping-for-cake-runner-for.net-framework). We should either provide OS redirects for every bootstrapper, or adopt docs to use PowerShell / Bash redirects. I'm also OK to update docs to move away from the OS redirects.

gep13:
Hmm, ok, fair enough.  In which case, this makes sense.

gep13 (approved):
LGTM!

PR #1221, opened 2020-11-08
Cleanup categories

pascalberger:
Cleanup categories for Cake.ArgumentBinder addin which are synced from nuget.org, but not really useful on Cake website


gep13:
If we remove these, won't they simply be put back next time the Addin Discoverer runs?

pascalberger:
> If we remove these, won't they simply be put back next time the Addin Discoverer runs?

No, categories are only synced when adding new addins. Later Addin Discoverer will only normalize line ending. See https://github.com/cake-contrib/Cake.AddinDiscoverer#what-is-automated

gep13 (approved):
LGTM

Issue #1078, opened 2020-10-07
Replace built-in / addin aliases category splitting by filtering on view


pascalberger:
Originally reported in https://github.com/cake-build/website/issues/244#issuecomment-285633049:

Currently categories on the reference page are splitted in "Built-In" and "Addins". As a user my main interested is to find an existing functionality, if it is in an addin or built-in most often is not my primary concern in this case.
Regarding this, wouldn't it be clearer to have just a single category listing in the left navigation sidebar, and then for each category three views rendered:

- One containing only built-in functionality
- One containing only functionality from addins
- One containing built-in functionality and functionality from addins

On the views a link/control could be placed to switch between the views (something like `[x] Show built-in [x] Show from addins`).

This would also solve the different problems existing with the current layout where built-in stuff is mixid with functionality from aliases.

pascalberger:
Related to https://github.com/cake-build/website/issues/927

pascalberger:
See also this comment: https://github.com/cake-build/website/issues/244#issuecomment-272660186

pascalberger:
Follow up to this issues to provide a switch to filter to only built-in aliases: https://github.com/cake-build/website/issues/1248

PR #1240, opened 2020-11-22
Show summary of aliases class

### LINKED ISSUE #1239: Add summary to addin detail page
We currently show summary on DSL category page (e.g. https://cakebuild.net/dsl/7zip/#Summary). This has some issues, like for example when the same category is used by different addins or by built-in aliases and extensions. Adding the summary to the extension detail page (e.g. https://cakebuild.net/extensions/cake-7zip/) would make more sense.

pascalberger:
Add summary of aliases class to addin details page. This is currently shown on the DSL page, but issue, there is that it doesn't work when multiple extensions or an extension and built-in alias use the same category. Having this documentation on the extension page is a better fit.

image

Updating DSL page is out of scope for this PR and will be done with #1078.

Fixes #1239

### DIFF
--- a/input/_ExtensionsLayout.cshtml
+++ b/input/_ExtensionsLayout.cshtml
@@ -49,6 +49,16 @@
                                     addinAssembly.Contains(
                                         aliasDocument.Document(CodeAnalysisKeys.ContainingAssembly).String("Name") + ".dll")));
 
+            // Show summary of aliases class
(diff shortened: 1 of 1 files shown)


gep13 (approved):
LGTM!

gep13:
@pascalberger thanks for fixing this up!

PR #1247, opened 2020-11-25
(GH-1078) Replace Built-In / Addin splitting in DSL category list 

### LINKED ISSUE #1078: Replace built-in / addin aliases category splitting by filtering on view

pascalberger:
Replace Built-In / Addin splitting in DSL category list with listing aliases for every category grouped by Built-In / Addin. This solves several issues like having summaries only shown for Built-In or one addin, when they share the same category, or having the annotation block shown that content is from a specific addin, while not everything on the page was from this addin.

image

Fixes #1078

### DIFF
--- a/input/_AddinAlert.cshtml
+++ b/input/_AddinAlert.cshtml
@@ -4,24 +4,23 @@
 	IDocument containingAssembly = Model.Get<IDocument>("ContainingAssembly");
     if(containingAssembly != null)
     {
-        IDocument addin = Documents["Extensions"]
-            .FirstOrDefault(x => 
+        IDocument extension = Documents["Extensions"]
+            .FirstOrDefault(x =>
                 (
                     x.String("Type") == null ||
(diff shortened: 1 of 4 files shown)


gep13 (approved):
LGTM!

gep13:
@pascalberger this looks amazing!

PR #1261, opened 2020-11-28
(GH-1107) nuget.cake: Ignore unlisted versions of packages

### LINKED ISSUE #1107: Ignore unlisted versions of packages
Website currently also reads unlisted versions of addin packages from nuget.org. An example is [Cake.Issues.Reporting.Generic](https://cakebuild.net/addins/issues/#cake-issues-reporting-generic) where the unlisted version [0.10.0-alpha0001](https://www.nuget.org/packages/Cake.Issues.Reporting.Generic/0.10.0-alpha0001) is used, instead of the latest listed version [0.9.0](https://www.nuget.org/packages/Cake.Issues.Reporting.Generic/0.9.0).

It should use the latest listed version.

augustoproiete:
Closes #1107

### DIFF
--- a/nuget.cake
+++ b/nuget.cake
@@ -1,8 +1,10 @@
 #addin "nuget:https://api.nuget.org/v3/index.json?package=Polly&version=7.1.0"
 #addin "nuget:https://api.nuget.org/v3/index.json?package=LitJson&version=0.13.0"
 #addin "nuget:https://api.nuget.org/v3/index.json??package=Cake.FileHelpers&version=3.3.0"
+#addin "nuget:https://api.nuget.org/v3/index.json?package=NuGet.Versioning&version=5.7.0"
 
 using System.Net.Http;
+using NuGet.Versioning;
 using Polly;
 
 public static void DownloadPackages(this ICakeContext context, DirectoryPath extensionDir, string[] packageIds)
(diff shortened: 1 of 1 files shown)


nuget.cake
```diff
@@ -36,14 +36,14 @@ public static void DownloadPackage(this ICakeContext context, DirectoryPath exte
...
+        where item?.catalogEntry?.listed ?? false
         select new
         {
             id = packageId,
-            version = packageItem?.upper,
+            version = item?.catalogEntry?.version,
             item?.packageContent
         }
-    ).FirstOrDefault();
+    ).LastOrDefault();
```
pascalberger, line 46: Shouldn't we order the list first to have a guaranteed order in case something changes on nuget API side?

pascalberger (commented):
@augustoproiete Thanks for fixing this. It seems to work, just one small comment

nuget.cake
augustoproiete: Cheers. I would be surprised if the NuGet team changed the sort order of this API as that would break so many tools... But I don't mind being a little conservative either. The only caveat is that we can't simply sort the list by the version string (as `0.10.0` would come before `0.2.0`, for example) so we'll have to use a natural string comparer or a parser that understands version numbers. I think using `NuGetVersion` from [`NuGet.Versioning`](https://www.nuget.org/packages/NuGet.Versioning/) might be the best/safest way to go. I'll update the PR shortly

augustoproiete: Sorting the list of NuGet packages by NuGetVersion via [`0c87ab3`](https://github.com/cake-build/website/pull/1261/commits/0c87ab3b23aadf6f12fb5c3c66dd6f57f621871c)

pascalberger (approved):
LGTM

pascalberger:

Issue #1291, opened 2020-12-02
Add explanation to supported version of extension


pascalberger:
We should add a tooltip to the supported version of the extension containing information what this number means. Something like:

```
These are the versions of Cake with an API compatible with the latest version of this addin. Depending on the API used in the extension it might also work with other versions of Cake.

Earlier version of the addin might be compatible with other versions of Cake. 
```

Fixes #1291

PR #1296, opened 2020-12-02
(GH-1291) Add explanation to supported version of extension

pascalberger:
Add explanation to supported Cake versions of an addin

### DIFF
--- a/input/_ExtensionsLayout.cshtml
+++ b/input/_ExtensionsLayout.cshtml
@@ -27,7 +27,7 @@
             @description
         </p>
 
-        @switch (type) { 
+        @switch (type) {
             case "Addin":
                 @Html.Partial("_UsageAddin");
                 break;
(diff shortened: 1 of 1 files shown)


input/_ExtensionsLayout.cshtml
```diff
@@ -81,6 +81,7 @@
             {
             <li>
                 Supported Cake versions: @supportedCakeVersions
+                <i class="fa fa-info-circle" data-toggle="tooltip" title="These are the versions of Cake with an API compatible with the latest version of this addin. Depending on the API used in the extension it might also work with other versions of Cake. Earlier version of the addin might be compatible with other versions of Cake."></i>
```
gep13, line 84:
```suggestion
                <i class="fa fa-info-circle" data-toggle="tooltip" title="These are the versions of Cake with an API compatible with the latest version of this addin. Depending on the API used in the extension it might also work with other versions of Cake. An earlier version of the addin might be compatible with other versions of Cake."></i>
```

gep13 (changes requested):

input/_ExtensionsLayout.cshtml
pascalberger: Sorry, saw just now that you created a suggestion. I added the change manually though.

gep13 (approved):
LGTM!

PR #1299, opened 2020-12-03
Add AnalyzedPackageVersion

### LINKED ISSUE #1298: Add AnalyzedPackageVersion
Add AnalyzedPackageVersion

Jericho:
This pull request was created by a tool: Cake.AddinDiscoverer version 3.57.0

Resolves #1298

### DIFF
--- a/extensions/Cake.AssemblyInfoReflector.yml
+++ b/extensions/Cake.AssemblyInfoReflector.yml
@@ -8,4 +8,5 @@ Author: wexman
 Description: Cake addin for retrieving assembly info from an existing assembly.
 Categories:
 - assembly info
-TargetCakeVersion: 0.26.1
+TargetCakeVersion: 0.18.0
+AnalyzedPackageVersion: 1.0.7
--- a/extensions/Cake.AutoRest.yml
+++ b/extensions/Cake.AutoRest.yml
@@ -9,4 +9,5 @@ Description: Cake addin for generating client code from compatible API definitio
 Categories:
 - azure
 - swagger
-TargetCakeVersion: 0.18.0
+TargetCakeVersion: 0.13.0
+AnalyzedPackageVersion: 0.1.2
--- a/extensions/Cake.Board.AzureBoards.yml
+++ b/extensions/Cake.Board.AzureBoards.yml
@@ -3,7 +3,7 @@ Name: Cake.Board.AzureBoards
 NuGet: Cake.Board.AzureBoards
 Assemblies:
 - "/**/Cake.Board.AzureBoards.dll"
-Repository: https://github.com/cake-contrib/Cake.Board/
+Repository: https://github.com/nicolabiancolini/Cake.Board/
 Author: Nicola Biancolini
 Description: This add-in allows you to interact directly with the work items of Azure Boards.
 Categories:
@@ -12,4 +12,5 @@ Categories:
 - work item
 - wit
 - board
-TargetCakeVersion: 0.34.1
+TargetCakeVersion: 0.33.0
+AnalyzedPackageVersion: 0.1.12
--- a/extensions/Cake.DocCreator.yml
+++ b/extensions/Cake.DocCreator.yml
@@ -3,10 +3,11 @@ Name: Cake.DocCreator
 NuGet: Cake.DocCreator
 Assemblies:
 - "/**/Cake.DocCreator.dll"
-Repository: https://github.com/agc93/DocCreator/
+Repository: https://bitbucket.org/agc93/DocCreator/
 Author: agc93
 Description: A Cake addin to generate HTML documentation from Markdown source, using DocCreator
 Categories:
 - documentation
 - markdown
-TargetCakeVersion: 0.19.4
+TargetCakeVersion: 0.10.1
+AnalyzedPackageVersion: 0.4.1
--- a/extensions/Cake.7zip.yml
+++ b/extensions/Cake.7zip.yml
@@ -9,3 +9,4 @@ Description: makes 7zip available as a tool in cake
 Categories:
 - sevenzip
 TargetCakeVersion: 0.33.0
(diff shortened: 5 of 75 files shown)


extensions/Cake.AssemblyInfoReflector.yml
```diff
@@ -8,4 +8,5 @@ Author: wexman
 Description: Cake addin for retrieving assembly info from an existing assembly.
 Categories:
 - assembly info
-TargetCakeVersion: 0.26.1
+TargetCakeVersion: 0.18.0
```
pascalberger, line 11: Why the change here?

extensions/Cake.AutoRest.yml
```diff
@@ -9,4 +9,5 @@ Description: Cake addin for generating client code from compatible API definitio
 Categories:
 - azure
 - swagger
-TargetCakeVersion: 0.18.0
+TargetCakeVersion: 0.13.0
```
pascalberger, line 12: Why the change here?

extensions/Cake.Board.AzureBoards.yml
```diff
@@ -3,7 +3,7 @@ Name: Cake.Board.AzureBoards
 NuGet: Cake.Board.AzureBoards
 Assemblies:
 - "/**/Cake.Board.AzureBoards.dll"
-Repository: https://github.com/cake-contrib/Cake.Board/
+Repository: https://github.com/nicolabiancolini/Cake.Board/
```
pascalberger, line 6: Why the change here?

extensions/Cake.Board.AzureBoards.yml
```diff
@@ -12,4 +12,5 @@ Categories:
 - work item
 - wit
 - board
-TargetCakeVersion: 0.34.1
+TargetCakeVersion: 0.33.0
```
pascalberger, line 15: Why the change here?

extensions/Cake.DocCreator.yml
```diff
@@ -3,10 +3,11 @@ Name: Cake.DocCreator
 NuGet: Cake.DocCreator
 Assemblies:
 - "/**/Cake.DocCreator.dll"
-Repository: https://github.com/agc93/DocCreator/
+Repository: https://bitbucket.org/agc93/DocCreator/
```
pascalberger, line 6: Why the change here?

extensions/Cake.DocCreator.yml
```diff
@@ -3,10 +3,11 @@ Name: Cake.DocCreator
...
 Author: agc93
 Description: A Cake addin to generate HTML documentation from Markdown source, using DocCreator
 Categories:
 - documentation
 - markdown
-TargetCakeVersion: 0.19.4
+TargetCakeVersion: 0.10.1
```
pascalberger, line 12: Why the change here?

pascalberger (changes requested):
@Jericho There were some changes beside the `AnalyzedPackageVersion` property being added, which seem wrong to me. What's the reason for these?

extensions/Cake.AssemblyInfoReflector.yml
Jericho: Because previously we were analyzing the most recent version of this addin (which is 1.1.0-beta2) but the logic was changed to use the most recent "stable" version which is an older version:

image

extensions/Cake.AutoRest.yml
Jericho: Because we changed the logic to ensure we analyze the most recent "stable" release:

image

extensions/Cake.DocCreator.yml
Jericho: Because we changed the logic to ensure we analyze the most recent "stable" release:

image

extensions/Cake.AssemblyInfoReflector.yml
pascalberger: Ah, didn't know you had this also already done in this version 👍

extensions/Cake.Board.AzureBoards.yml
Jericho: I don't know for sure. I'm guessing that's what was in the nuspec in the older package. Give me a few minutes to research...

pascalberger (approved):

extensions/Cake.Board.AzureBoards.yml
Jericho: Ok I figured it out. It's a little complicated:
1. the nuspec of the most recent **prerelease** package contains URL which points to the repo under the cake-contrib organization. This explains the `https://github.com/cake-contrib/Cake.Board/` value which was previously in the YAML
2. the nuspec in the older "stable" package contains the URL which points to the original author's repo (presumably because at that time the repo had not been transferred to the cake-contrib organization yet)
3. AddinDiscoverer has logic to check if an addin has been transferred to the cake-contrib organization and if so, automatically update the URL. In this case however, this logic fails because the name of the repo does not match the name of the addin: the repo is "Cake.Board" and the addin is "Cake.Board.AzureBoards".

extensions/Cake.DocCreator.yml
Jericho: I haven't done the research but my guess is that the author moved the repo from bitbucket to github between June 2016 and April 2017. That would explain the difference between these two URLs.

extensions/Cake.Board.AzureBoards.yml
Jericho: So in conclusion, the different values are explained because the nuspec changed between July 2019 and September 2019 AND ALSO AddinDisco was unable to automatically fix the URL because of the discrepancy between the repo name and the addin name.

Make sense?

pascalberger: Yes, makes sense

PR #1311, opened 2020-12-07
Update configuration documentation

pascalberger:
Orders config parameters alphabetically, uses tab control to show different ways of setting config and makes use of annotations.


gep13 (approved):
LGTM!

PR #1359, opened 2020-12-14
GH-1153: Show aliases without ICakeContext extension parameter

### LINKED ISSUE #1153: Show aliases without extension parameter
When listing aliases we currently show the syntax of the extension method. 

image

That they are extensions is an implementation detail from a user point of view, and we therefore should show the syntax as the user will call it (without the `ICakeContext` parameter)

augustoproiete:
image

image

---

Closes #1153


pascalberger (approved):
LGTM

pascalberger:

Issue #580, opened 2018-10-02
Show target .NET framework for an addin


patriksvensson:
Would be nice if we could show some addin requirements on the addin pages.

* .NET/Mono only?
* Windows only? (Not sure if this is doable)

Not sure what is needed for this to work, but just showing whether or not an addin is .NET/Mono only would be good. Might act as an incentive for people to multitarget as well.

Assigning @daveaglick since he explicitly asked for it 😄

pascalberger:
Would also be nice to see against which version of Cake the addin was built (and it therefore requires)

pascalberger:
> Would also be nice to see against which version of Cake the addin was built (and it therefore requires)

This is https://github.com/cake-build/website/issues/299

augustoproiete:
I'm thinking about how one would implement this one.

> ### Show required .NET version
> - .NET/Mono only?

Would it make sense to display a list with all the .NET Target Frameworks of the add-in, on the right sidebar? Something along those lines:

![image](https://user-images.githubusercontent.com/177608/100534360-44b5c600-31e4-11eb-8b0e-fe2f797852b7.png)

I think TFM would be a good fit as it's shorter and devs should be familiar with it by now, but we could also display a more friendly name with target framework + version (e.g. .NET Standard 2.1).

---

> ### Windows only?

I don't think there's a way for us to determine *with certainty* if an add-in runs on Windows and/or Linux and/or OSX. I suggest we leave that for the add-in author to tell us by introducing a new attribute (or set of attributes) in `Cake.Core.Annotations` that the add-in author can apply at the assembly level to provide more information about the operating system the Cake add-in supports.

e.g.

```
[assembly: CakeSupportedOS("Linux")]
[assembly: CakeSupportedOS("MacOSX")]
[assembly: CakeSupportedOS("Windows")]
```

Is the above enough? Would it make more sense to be more specific with OS minimum version e.g. `Windows7`?

_It might make sense to move the this last part to a separate issue and tackle the Target Frameworks independently first_.

pascalberger:
Original issue is already a few years old and some things have changed since then. I would suggest the following:

1. Show TFM on extension detail page (like you suggested)
2. Show compatible runners on extension detail page (this is probably what most users are interested in)
3. Add a filter on the extension overview page to filter to extensions which are available for specific runners

augustoproiete:
Following up on this one:

> 1. Show TFM on extension detail page (like you suggested)

[...]

PR #1361, opened 2020-12-14
(GH-580) Add target framework versions to extension metadata

pascalberger:
Add support for target framework versions to extension metadata.

Part of #580

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -48,6 +48,7 @@ class ExtensionSpec
     public List<string> Categories { get; set; }
     public string AnalyzedPackageVersion { get; set; }
     public string TargetCakeVersion { get; set; }
+    public List<string> TargetFrameworks { get; set; }
 }
 
 // Variables


build.cake
```diff
@@ -48,6 +48,7 @@ class ExtensionSpec
     public List<string> Categories { get; set; }
     public string AnalyzedPackageVersion { get; set; }
     public string TargetCakeVersion { get; set; }
+    public List<string> TargetFrameworkVersions { get; set; }
```
Jericho, line 51: Today I learned that TFM stands for "Target Framework Moniker" ([here](https://docs.microsoft.com/en-us/dotnet/standard/frameworks)). So how about we name this property `TargetFrameworkMonikers`?

augustoproiete (approved):

Jericho (changes requested):
Here's a suggestion:

build.cake
augustoproiete: `TargetFrameworkMonikers` works for me, as well as `TargetFrameworks` which is what holds the TFMs in the build.

```
  <PropertyGroup>
    <TargetFrameworks>net5.0;netstandard2.0;net461</TargetFrameworks>
  </PropertyGroup>
```

Jericho: `TargetFrameworks` is shorter, I like it.

pascalberger: Works both for me. How about going with `TargetFrameworks`?

Jericho: I vote for `TargetFrameworks`. It's shorter, it's similar to the XML node in.csproj files. It's better than my suggestion.

pascalberger: Updated to `TargetFrameworks`

Jericho (approved):
Looks good!

PR #1362, opened 2020-12-15
GH-1153: Show aliases without ICakeContext extension parameter (also when it's the only arg)

augustoproiete:
I missed a case in PR #1359 where `ICakeContext` is the only argument of the method - and in that case it still shows in the documentation.

As an example, the [Azure DevOps add-in](https://cakebuild.net/dsl/azure-devops/) at the moment shows:

image

---

But should be:

image


pascalberger (approved):

pascalberger:

Issue #1367, opened 2020-12-16
Add release date of extensions


pascalberger:
Since we now have version specific information in the extension YAML files it would be nice to also have the release date of the `AnalyzedPackageVersion`.

pascalberger:
Related issue for addin discoverer: https://github.com/cake-contrib/Cake.AddinDiscoverer/issues/159

PR #1383, opened 2020-12-19
Update display of repository URL

pascalberger:
Hides the repo URL from the extension list in preparation for https://github.com/cake-build/website/issues/1367 which provides update date for extensions, which is a much more interesting information in a list of extensions. 
Also handles the case where an addin don't has a repository URL and removes the `.git` suffix from the repo link name.


gep13 (approved):
LGTM

PR #1394, opened 2020-12-20
Use fix titles for links

pascalberger:
Use fix titles for repo and project URL instead of showing the URL itself. This solves issues when the URL is longer than the available space, e.g. here:

image

After this PR:

image


gep13 (approved):
LGTM!

gep13:
This makes a lot of sense to me, thanks for doing this!

PR #1397, opened 2020-12-20
Show project URL only if different from repo URL

pascalberger:
Show project URL only if different from repo URL since it doesn't provide any additional value to the user if we show a project url which is the also the repo URL


gep13 (approved):
LGTM!

PR #1266, opened 2020-11-29
(GH-184) Display info when add-in is pre-release & fix preprocessor directive

augustoproiete:
- [x] Display information alert when the add-in is a pre-release version
- [x] Include the `&prerelease` parameter in the preprocessor directive example

image

---

Closes #184

PR #1399, opened 2020-12-20
GH-1294: Leverage AnalyzedPackageVersion and AnalyzedPackageIsPrerelease from addin metadata

### LINKED ISSUE #1294: Define version to use for every extension in YML file
Since we started to have version dependent data (`TargetCakeVersion`) in the extension YML file, it makes sense to also have the version from which this data comes added. This version should be used on the website for the specific extension.  

Corresponding issue for addin discoverer: https://github.com/cake-contrib/Cake.AddinDiscoverer/issues/143

augustoproiete:
- Download addin nuget packages using `AnalyzedPackageVersion` (removed logic of downloading latest available)
- Use `AnalyzedPackageIsPrerelease` to determine if package is prerelease (for [info alert and preprocessor directive](https://github.com/cake-build/website/pull/1266))

---

Closes #1294

### DIFF
--- a/nuget.cake
+++ b/nuget.cake
@@ -1,5 +1,4 @@
 #addin "nuget:https://api.nuget.org/v3/index.json?package=Polly&version=7.1.0"
-#addin "nuget:https://api.nuget.org/v3/index.json?package=LitJson&version=0.13.0"
 #addin "nuget:https://api.nuget.org/v3/index.json??package=Cake.FileHelpers&version=3.3.0"
 #addin "nuget:https://api.nuget.org/v3/index.json?package=NuGet.Protocol&version=5.7.0"
 
@@ -12,7 +11,7 @@ using NuGetRepository = NuGet.Protocol.Core.Types.Repository;
 using NuGet.Versioning;
 using Polly;
 
-public static void DownloadPackages(this ICakeContext context, DirectoryPath extensionDir, string[] packageIds)
+public static void DownloadPackages(this ICakeContext context, DirectoryPath extensionDir, IDictionary<string, string> packageIdAndVersionLookup)
 {
     var retryPolicy = Policy
                         .Handle<Exception>()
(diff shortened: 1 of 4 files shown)


nuget.cake
```diff
@@ -23,45 +22,36 @@ public static void DownloadPackages(this ICakeContext context, DirectoryPath ext
...
-            item?.packageContent
-        }
-    ).LastOrDefault();
+    var packageId = packageIdAndVersion.Key;
+    var packageVersion = packageIdAndVersion.Value;
 
-    context.Verbose("[{0}] found {1}...", packageId, packageInfo);
+    if (string.IsNullOrWhiteSpace(packageVersion))
+    {
+        context.Warning($"[{{0}}] Skipping download of {packageId} - AnalyzedPackageVersion not specified in metadata", packageId);
```
pascalberger, line 40: This will skip downloading [Cake.Issues.Testing](https://github.com/cake-build/website/blob/master/extensions/Cake.Issues.Testing.yml) which doesn't have a `AnalyzedPackageVersion` in the metadata, since it is not maintained by Addin Discoverer (since it is not an extension, but helpers for writing tests). Idea was to have it included in API documentation on the website. The question is if we want to get rid of it and only handle extensions, which can be updated by  Addin Discoverer, or if we should fallback to download latest version in these cases.

augustoproiete: `Cake.Issues.Testing` seems to be the only exception so far. Would it make sense to manually add `AnalyzedPackageVersion` to the `Cake.Issues.Testing` YAML for now, until Addin Discoverer is updated to handle these cases as well?

I don't know the history behind why it skips this one, but I had the impression the goal of Addin Discoverer is to eventually handle all kinds of extensions to Cake, be it modules, or APIs like this one.

pascalberger: `Cake.Issues.Testing` is special since it is a NuGet package which helps to write tests for addins built on top of the Cake.Issues addin. It's neither an addin, nor a module or recipe script. It's also not shown in the extension list of the website, just the api part. Before we start to implement additional logic for it I think I just delete it to make stuff simpler.

pascalberger (approved):

pascalberger:

Issue #1403, opened 2020-12-21
Show supported runners on extension detail page


pascalberger:
Follow up of https://github.com/cake-build/website/issues/580 and https://github.com/cake-build/website/issues/299. With the combination of Cake version the addin references and TFM we can decide with which runners it is compatible with based on the [supported platform table](https://cakebuild.net/docs/running-builds/runners/#supported-platforms). 
An addin should be classified as compatible with a runner as soon as it targets a TFM, which is compatible with a runner. It shouldn't matter if an addin targets `netstandard2.0` or e.g. multi-targets `net641` and `netcoreapp3.0`.

This is of course only from a technical point of view, it's still possible that e.g. an addin targeting netstandard2.0 doesn't work on Cake .NET Tool on Linux. We should mention this in a tooltip.

pascalberger:
This probably can be done in Addin Discoverer as part of the compliance score calculation (https://github.com/cake-contrib/Cake.AddinDiscoverer/issues/156) and written to YAML file

PR #1404, opened 2020-12-21
Add separate section for compatibility with Cake runners

pascalberger:
Supported Cake versions should have the same weight as Target Framework versions. This PR therefore moves the supported Cake versions into their own section in the side-bar. 

This is also in preparation for https://github.com/cake-build/website/issues/1403 where we can also list compatible runners in this section beside compatible versions.

image


augustoproiete (approved):

gep13 (approved):
LGTM!

PR #1406, opened 2020-12-21
Add documentation for bootstrapping CLI switches

pascalberger:
Add documentation for bootstrapping CLI switches

### DIFF
--- a/input/Shared/switches.txt
+++ b/input/Shared/switches.txt
@@ -1,24 +1,28 @@
 ## Switches
 
-| Switch                    | Description                                                                                        | Available Since |
-|---------------------------|----------------------------------------------------------------------------------------------------|-----------------|
-| --verbosity=&lt;value&gt; | Specifies the amount of information to be displayed (quiet, minimal, normal, verbose, diagnostic). | [0.1.1]         |
-| --debug                   | Launches script in debug mode.                                                                     | [0.12.0]        |
-| --showdescription         | Shows description about tasks.                                                                     | [0.1.10]        |
-| --showtree, --tree        | Shows the task dependency tree.                                                                    | [0.31.0]        |
-| --dryrun                  | Performs a dry run.                                                                                | [0.2.0]         |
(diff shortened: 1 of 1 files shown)


input/Shared/switches.txt
```diff
@@ -8,6 +8,8 @@
 | --showtree, --tree        | Shows the task dependency tree.                                                                    | [0.31.0]        |
 | --dryrun                  | Performs a dry run.                                                                                | [0.2.0]         |
 | --exclusive               | Execute a single task without any dependencies.                                                    | [0.29.0]        |
+| --bootstrapping           | Download/install modules defined by `#module` directives, but do not run build.                    | [0.24.0]        |
```
gep13, line 11: Isn't this just `bootstrap`?

gep13 (changes requested):

input/Shared/switches.txt
pascalberger: Of course. Fixed

gep13: Also, I think I left this out of the original list that I put together, due to the fact that this is being removed, unless we add another column to say when it was removed.

pascalberger: It hasen't been removed

gep13: Why is it needed?  This happens automatically now?

pascalberger: I was not really involved with this, just stumbled across this as I tried to update CLI switch documentation. Explicit bootstrapping was done with https://github.com/cake-build/cake/issues/2833. Maybe because we didn't want to break stuff?  @patriksvensson or @devlead might have more details.

pascalberger: As discussed with @devlead :

> To not break every existing bootstrapper script/github action/azure pipelines extension etc.
It still works and does what it did before.
If we remove it now it would potentially mean scripts execute twice and that would be bad.

gep13 (approved):
LGTM!

PR #1443, opened 2020-12-25
Fix examples for Cake Frosting

pascalberger:
Fix examples for Cake Frosting in the Set Configuration Values page


gep13 (approved):
LGTM!

PR #1452, opened 2020-12-26
Only show arguments in examples

pascalberger:
We currently have different documentation how to run the different runners.
- http://cakebuild.net/docs/writing-builds/running-targets#passing-a-target-to-the-script uses `dotnet cake` for tool and bootstrapper for other runners
- http://cakebuild.net/docs/writing-builds/args-and-environment-vars#passing-and-reading-arguments is without bootstrapper for all runners
- https://cakebuild.net/docs/running-builds/configuration/set-configuration-values uses bootstrapper for Cake Frosting, but not for other runners.

This PR makes examples consistent by removing the Cake invocation part and only showing the arguments. Instead of showing call with bootstrapper (which is still different for every OS this makes documentation much more focused on the specific topic which we want to explain. Different ways how to run Cake, with or without bootstrapper and for every different runner is available on documentation page for every runner (e.g. https://cakebuild.net/docs/running-builds/runners/dotnet-tool).

Also removes the example in Set Configuration Values where we show how to pass multiple NuGet sources on command line, since this is something specific for this configuration value and not specific to setting configuratoin values on command line and already documented at https://cakebuild.net/docs/running-builds/configuration/default-configuration-values#nuget-download-url.

### DIFF
--- a/input/docs/running-builds/configuration/set-configuration-values.md
+++ b/input/docs/running-builds/configuration/set-configuration-values.md
@@ -46,82 +46,8 @@ The configuration file should be located in the same directory as your Cake scri
 
 Finally, you can specify an input parameter directly to Cake, in the following format:
 
-<ul class="nav nav-tabs">
-    <li class="active"><a data-toggle="tab" href="#tool1">Cake .NET Tool</a></li>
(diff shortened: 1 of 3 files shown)


gep13 (approved):
LGTM

Issue #1139, opened 2020-10-14
Show latest version on homepage


pascalberger:
It would be nice to have latest version shown on landing page with a link to the [runners overview](https://cakebuild.net/docs/running-builds/runners/)

chrysillala:
Hi, I would like to work on this as part of Hacktoberfest. Could you please assign and guide me? Thanks!

pascalberger:
@chrysillala Thanks for giving this a try.

My idea is to have something like this, where the text `Version 0.38.5 is now available` should be a link to the release blog post.

![image](https://user-images.githubusercontent.com/2190718/96353037-c801d900-10c8-11eb-96c4-05bcaf9baf8f.png)

The website is a static website built using [Wyam](https://wyam.io). We don't build website scheduled (e.g. once a day), only when changes are done. 

I see two options how to get the latest version:

1. From GitHub.

    In case of a release we will first publish GitHub release before publishing a blog post, therefore it should be possible to retrieve latest version of Cake through GitHub API (https://api.github.com/repos/cake-build/cake/releases?per_page=1) as part of the Wyam pipeline. There's a [Wyam GitHub module](https://wyam.io/modules/github) available which probably could be used to call the API. 

2. From Blog Post

    Release blog posts are in a separate category: https://cakebuild.net/blog/Release-Notes/. It should be possible to get the newest and release version from it. It might also be possible to add additional metadata to the blog posts to easier get the release name. Advantage of this approach is that you already would have the link to the blog post. The downside of this approach is that it would require a blog post for every release. Stuff like [this blog post](https://cakebuild.net/blog/2020/06/cake-v0.38.3-released-in-v0.38.3) would probably break stuff.

chrysillala:
@pascalberger Thanks for the detailed explanation. Will keep you updated.

chrysillala:
Hi @pascalberger, I was following the `getting-started` guide and then stuck at this step `dotnet tool install Cake.Tool --version x.y.z`. 
This is the error message: `Specified version 'x.y.z' is not a valid NuGet version range.`
Could you please help me? Thanks!

gep13:
x.y.z is shown as an indication that you need to provide a version number. The current latest release version is 0.38.5.

pascalberger:
@chrysillala What are you trying to achieve? The getting started guide is if you want to create a new project using Cake. You'll need to replace `x.y.z` with the version number of Cake you want to use in this case.

[...]

Issue #1428, opened 2020-12-23
Automatically use latest version in getting started guide


pascalberger:
In getting started guides (e.g. https://cakebuild.net/docs/getting-started/setting-up-a-new-project) we use a placeholder for installing Cake:

```
dotnet tool install Cake.Tool --version x.y.z
``` 

It would be helpful if we can show the example with latest stable version by default.

pascalberger:
Can probably be done as a follow up of https://github.com/cake-build/website/issues/1139

PR #1459, opened 2020-12-30
Show latest Cake version on home page

augustoproiete:
image

Given that the build pipeline already uses the GitHub API to get the latest release (and download the source) and also the fact that we expect to use the latest version number in multiple places (e.g. https://github.com/cake-build/website/issues/1139, https://github.com/cake-build/website/issues/1428) I decided on the following approach:

- Use the GitHub API to get latest release from cake-build/cake repo via the Cake build
- Add it to the global Wyam metadata via settings
- Search for a blog post with a matching `releaseName` on the front matter (1)
    - Link to the blog post if found, otherwise link to the GitHub release

(1) This means that moving forward, posts of release announcements should include a metadata `releaseName` in the front matter with the same name of the GitHub release (e.g. v1.0.0-rc0002) for the alert on the home page to automatically link to the blog post. In case there is no blog post, the alert will link to the release notes on GitHub.

---

Closes https://github.com/cake-build/website/issues/1139

### DIFF
--- a/github.cake
+++ b/github.cake
@@ -1,4 +1,4 @@
-#addin "nuget:https://api.nuget.org/v3/index.json?package=Octokit&version=0.32.0"
+#addin "nuget:https://api.nuget.org/v3/index.json?package=Octokit&version=0.48.0"
 #addin "nuget:https://api.nuget.org/v3/index.json?package=NuGet.Versioning&version=5.7.0"
 
 using Octokit;
@@ -51,6 +51,7 @@ public static CakeGitHubReleaseInfo GetCakeGitHubReleaseInfo(this ICakeContext c
     {
         LatestReleaseName = latestCakeRelease.Name,
         LatestReleaseUrl = latestCakeRelease.HtmlUrl,
+        LatestReleaseZipUrl = latestCakeRelease.ZipballUrl,
     };
 
     LogGitHubReleaseInfo(context, _cakeGitHubReleaseInfo);
@@ -65,10 +66,12 @@ private static void LogGitHubReleaseInfo(ICakeContext context, CakeGitHubRelease
 {
     context.Information("Cake Latest Release Name: {0}", releaseInfo.LatestReleaseName);
     context.Information("Cake Latest Release Url: {0}", releaseInfo.LatestReleaseUrl);
+    context.Information("Cake Latest Release Zip Url: {0}", releaseInfo.LatestReleaseZipUrl);
 }
 
 public class CakeGitHubReleaseInfo
 {
     public string LatestReleaseName { get; set; }
     public string LatestReleaseUrl { get; set; }
+    public string LatestReleaseZipUrl { get; set; }
 }
--- a/github.cake
+++ b/github.cake
@@ -1,0 +1,74 @@
+#addin "nuget:https://api.nuget.org/v3/index.json?package=Octokit&version=0.32.0"
+#addin "nuget:https://api.nuget.org/v3/index.json?package=NuGet.Versioning&version=5.7.0"
+
+using Octokit;
+using NuGet.Versioning;
+
+static CakeGitHubReleaseInfo _cakeGitHubReleaseInfo;
+
+public static CakeGitHubReleaseInfo GetCakeGitHubReleaseInfo(this ICakeContext context, DirectoryPath releaseDir)
+{
+    if (!(_cakeGitHubReleaseInfo is null))
+    {
+        return _cakeGitHubReleaseInfo;
+    }
+
+    // We cache the latest Cake release information in a file to speed up local development and also
+    // to decrease the chance of getting rate limited by GitHub when running unauthenticated
+    var cachedGitHubDir = context.MakeAbsolute(releaseDir.Combine("cake-github"));
+    var cachedReleaseInfoFile = cachedGitHubDir.CombineWithFilePath($"{nameof(CakeGitHubReleaseInfo)}.yml");
+    if (context.FileExists(cachedReleaseInfoFile))
+    {
+        context.Verbose("Retrieving Cake release information from cached file {0}...", cachedReleaseInfoFile);
+
+        _cakeGitHubReleaseInfo = context.DeserializeYamlFromFile<CakeGitHubReleaseInfo>(cachedReleaseInfoFile);
+        LogGitHubReleaseInfo(context, _cakeGitHubReleaseInfo);
+
+        return _cakeGitHubReleaseInfo;
+    }
+
+    context.Information("Retrieving Cake release information from GitHub...");
+
+    var client = new GitHubClient(new ProductHeaderValue("cake-build-website"));
+
+    var gitHubAccessToken =  context.EnvironmentVariable("git_access_token");
+    if (!string.IsNullOrWhiteSpace(gitHubAccessToken))
+    {
+        client.Credentials = new Credentials(gitHubAccessToken);
+    }
+
+    var allCakeReleases = client.Repository.Release.GetAll("cake-build", "cake")
+        .GetAwaiter()
+        .GetResult()
+        .Where(r => !r.Draft && r.PublishedAt.HasValue && r.Name.StartsWith("v", StringComparison.Ordinal))
+        .OrderByDescending(r => new NuGetVersion(r.Name.Substring(1)))
+        .ToList();
+
+    var latestCakeRelease = allCakeReleases
+        .First();
+
+    _cakeGitHubReleaseInfo = new CakeGitHubReleaseInfo
+    {
+        LatestReleaseName = latestCakeRelease.Name,
+        LatestReleaseUrl = latestCakeRelease.HtmlUrl,
+    };
+
+    LogGitHubReleaseInfo(context, _cakeGitHubReleaseInfo);
+
+    context.EnsureDirectoryExists(cachedReleaseInfoFile.GetDirectory());
+    context.SerializeYamlToFile<CakeGitHubReleaseInfo>(cachedReleaseInfoFile, _cakeGitHubReleaseInfo);
+
+    return _cakeGitHubReleaseInfo;
+}
+
+private static void LogGitHubReleaseInfo(ICakeContext context, CakeGitHubReleaseInfo releaseInfo)
+{
+    context.Information("Cake Latest Release Name: {0}", releaseInfo.LatestReleaseName);
+    context.Information("Cake Latest Release Url: {0}", releaseInfo.LatestReleaseUrl);
+}
+
+public class CakeGitHubReleaseInfo
+{
+    public string LatestReleaseName { get; set; }
+    public string LatestReleaseUrl { get; set; }
+}
--- a/input/index.cshtml
+++ b/input/index.cshtml
@@ -3,6 +3,28 @@ NoSidebar: true
 NoContainer: true
 NoGutter: true
 ---
+@{
+    // Latest Cake release name from GitHub e.g. v1.0.0-rc0002
+    var latestReleaseName = Model.String("CakeLatestReleaseName") ?? "x.y.z";
+
+    // All documents in the "Release Notes" category
+    var releaseBlogPosts = Documents
+        .Where(d => "Release Notes".Equals(d.String("category"), StringComparison.Ordinal));
+
+    // Attempt to locate a corresponding blog post for the latest Cake release
+    // NB: Blog posts need to have `releaseName` in the front matter matching the GitHub release name
+    var latestReleaseBlogPost = releaseBlogPosts
+        .Where(p => latestReleaseName.Equals(p.String("releaseName"), StringComparison.Ordinal))
+        .FirstOrDefault();
+
+    // Use link to blog post if found, otherwise fallback to link to GitHub release
+    var latestReleaseUrl = latestReleaseBlogPost is null
+        ? Model.String("CakeLatestReleaseUrl") ?? "#"
+        : Context.GetLink(latestReleaseBlogPost);
+}
+<div class="alert-latest-version">
+    <a href="@latestReleaseUrl">Cake @latestReleaseName is now available</a>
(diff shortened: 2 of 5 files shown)


github.cake
```diff
@@ -0,0 +1,78 @@
+#addin "nuget:https://api.nuget.org/v3/index.json?package=Octokit&version=0.48.0"
+#addin "nuget:https://api.nuget.org/v3/index.json?package=NuGet.Versioning&version=5.7.0"
+#addin "nuget:https://api.nuget.org/v3/index.json?package=Cake.Yaml&version=3.1.1"
```
pascalberger, line 3: Cake.Yaml addin has already been added in the  `build.cake` file. Addins should only be added once, otherwise it could lead to a lot of confusion if different versions are used. I would add them all to the `build.cake` file to have it in a single place (or in a separate `addins.cake` file).

github.cake
```diff
@@ -0,0 +1,78 @@
...
+
+using Octokit;
+using NuGet.Versioning;
+
+static CakeGitHubReleaseInfo _cakeGitHubReleaseInfo;
```
pascalberger, line 8: Instead of using a global variable, we can use a [typed context](https://cakebuild.net/docs/writing-builds/sharing-build-state#typed-context) to share release information across tasks.

github.cake
```diff
@@ -0,0 +1,78 @@
...
+
+public static CakeGitHubReleaseInfo GetCakeGitHubReleaseInfo(this ICakeContext context, DirectoryPath releaseDir)
+{
+    if (!(_cakeGitHubReleaseInfo is null))
+    {
+        return _cakeGitHubReleaseInfo;
+    }
+
+    // We cache the latest Cake release information in a file to speed up local development and also
+    // to decrease the chance of getting rate limited by GitHub when running unauthenticated
```
pascalberger, line 18: Instead of caching we can put the reading of GitHub release information into a separate task, which is not called from the `preview` target, but a dependency of the `GetArtifacts` tasks, so that it is called in the `Build` / `Default` target. This is how it is already done with reading of Cake source code and extension packages.

github.cake
```diff
@@ -0,0 +1,78 @@
...
+        LogGitHubReleaseInfo(context, _cakeGitHubReleaseInfo);
+
+        return _cakeGitHubReleaseInfo;
+    }
+
+    context.Information("Retrieving Cake release information from GitHub...");
+
+    var client = new GitHubClient(new ProductHeaderValue("cake-build-website"));
+
+    var gitHubAccessToken =  context.EnvironmentVariable("git_access_token");
```
pascalberger, line 35: You can use the global `accessToken` variable which is set in the `build.cake` file here

github.cake
```diff
@@ -0,0 +1,78 @@
...
+    if (!string.IsNullOrWhiteSpace(gitHubAccessToken))
+    {
+        client.Credentials = new Credentials(gitHubAccessToken);
+    }
+
+    var allCakeReleases = client.Repository.Release.GetAll("cake-build", "cake")
```
pascalberger, line 41: Can we use `client.Repository.Release.GetLatest("cake-build", "cake").Result` here? I think this will ignore pre-release versions. But the question is if we want to show pre-release versions or not? For the banner on start page lates pre-release version probably makes most sense.

input/index.cshtml
```diff
@@ -3,6 +3,28 @@ NoSidebar: true
 NoContainer: true
 NoGutter: true
 ---
+@{
+    // Latest Cake release name from GitHub e.g. v1.0.0-rc0002
+    var latestReleaseName = Model.String("CakeLatestReleaseName") ?? "x.y.z";
```
pascalberger, line 8: `x.y.z` doesn't seem like a useful fallback. Wouldn't just hide the banner be a better option if no information is passed?

pascalberger (changes requested):
@augustoproiete Many thanks for doing this 👍 

I've added a few comments. Mostly for implementation details.

github.cake
augustoproiete: Interesting. I just followed the prior art of [nuget.cake](https://github.com/cake-build/website/blob/79122cbfde83fa81334a7e3576c0acb435137cec/nuget.cake#L1-L3). Should we do the same with that one, then?

One nice thing about having the addins declared in the separate Cake file is the ability to run it directly (via dotnet cake) without having to run any task in build.cake, which is useful during development/testing.

If we had an `addins.cake` as you suggest, that is included by all the other files i.e. `build.cake`, `nuget.cake`, and `github.cake`, would that be a better approach? I'm assuming Cake would be smart enough to include the file only once, but haven't tested it does.

pascalberger: `nuget.cake` only added addins, there are no duplications between `nuget.cake` and `build.cake`. The main issue is that addin is now added twice. The different `*.cake` files are basically just concatenated to a single file before they are processed by Roslyn. It's not that they have a separate scope like e.g. classes. If an addin is already loaded in `build.cake` it is also available in every other `.cake` file. The addins can be added to different files, but should not be added twice. At a certain complexity I personally find it easier to maintain if they are all in a single place (either the main cake file or a separate file). If we add an `addin.cake` it just needs to be loaded once in `build.cake`.

github.cake
augustoproiete: I'm fine with reading the GitHub release information in a separate task, but we still need to get the release version info during the Preview Task as well in order to see the banner in the docs, and in the getting started, thus the Preview task would also have to depend on this new Task and this is where caching is important given the Preview task can be run a number of times during local dev. I remember hitting the limit a couple of times yesterday - I believe the limit is 30 unauthenticated requests.

github.cake
augustoproiete: My intent was to move everything GitHub-related to the `github.cake` file, thus I believe this variable no longer exists in `build.cake` - unless I forgot to delete, in which case I will.

github.cake
augustoproiete: Indeed `GetLatest` ignores prerelease versions (and thus returns v0.38.5 as of this writing) so it would not be an option if we want to show the latest prerelease in the banner (and link to the prerelease blog post).

The prerelease version is also the one where the source is downloaded from at the moment (even before this PR), thus all the API docs are using the prerelease assemblies, so I think it makes sense to promote the version that appears in the docs.

input/index.cshtml
augustoproiete: The `x.y.z` should never really be hit unless you comment something out in `build.cake` / `github.cake` during local development, so it's just a local dev fallback to see something in the banner.

github.cake
augustoproiete: Is there a rule of thumb for deciding between putting code in a build task or in a utility/extension to `ICakeContext`? I followed the prior art of [nuget.cake](https://github.com/cake-build/website/blob/79122cbfde83fa81334a7e3576c0acb435137cec/nuget.cake) with adding an extension method to `ICakeContext`, so the in-memory cache is an implementation detail of the extension method itself and not a shared context data of the build (i.e. the method returns everything that the task that called it needs, similar to the nuget.cake extensions.

github.cake
augustoproiete: Indeed I forgot to delete the `accessToken` variable from `build.cake`. I deleted it just now and updated the PR. My thinking here is that the dependency is top-down i.e. `build.cake` depends on `github.cake` so it knows what to expect from it, whereas `github.cake` doesn't have any knowledge of who is including them, so expecting a global variable `accessToken` seems like an obscure dependency for `github.cake` to have.

That said, if the goal of the separate `.cake` files is _not_ to have reusable code across `.cake` files, but just make `build.cake` smaller, perhaps it would make sense to rename the other files to something like `build._nuget.cake` & `build._github.cake` to make the relationship more explicit (with the underscore hinting it's a private dependency).

github.cake
pascalberger: I was talking about the global `_cakeGitHubReleaseInfo` variable. I assume the reason why this is a global variable is because the method is called twice, to avoid having to call out twice to GitHub. Instead of a global variable, the method can be called once and the result be stored in a typed context which then can be passed across the tasks. But I'm OK to keep it like this for the moment. We can look at refactoring the build process in a later step (especially since we didn't use any typed context until now)

github.cake
pascalberger: If the task writes the result to a file, like the caching implementation currently does, it would work. It would then be in the responsibility of the developer to call the `GetArtifacts` task once to get the release information, otherwise you won't have this information. This is how it currently is also implemented for extensions or Cake Source. If I don't call `GetArtifacts` before calling `Preview` I won't have this information in the built website (and in most of the cases this is even the better, because faster option).

github.cake
pascalberger: > My thinking here is that the dependency is top-down i.e. `build.cake` depends on `github.cake` so it knows what to expect from it, whereas `github.cake` doesn't have any knowledge of who is including them, so expecting a global variable `accessToken` seems like an obscure dependency for `github.cake` to have.
>
> That said, if the goal of the separate `.cake` files is _not_ to have reusable code across `.cake` files, but just make `build.cake` smaller, perhaps it would make sense to rename the other files to something like `build._nuget.cake` & `build._github.cake` to make the relationship more explicit (with the underscore hinting it's a private dependency).

You should not confuse Cake scripts with e.g. classes. It's really as simple as the scripts mentioned with `#load` are concatenated together in one big file. There's no separation or scoping going on here. A global variable, regardless in which script it is declared, will be available in any script since in the end it will be concatenated into one single file. So the reason fo splitting into multiple `.cake` files  is only to make files smaller and code better readable.  To decouple dependencies of the code, keeping state in a typed context would be the best way, since with this state can be passed across tasks and can also into helper methods. There won't be any global variables in this case and therefore also no more dependency.

But same as in the comment above. I'm OK with merging it like this and refactoring of build or introduction of a typed context can be done in a later PR.

github.cake
pascalberger: Agree that it is correct for now. But I think I'll create an issue that we should think about which version we want to document in the API (IMHO for API documentation latest stable would make sense, while for the news banner on the home page I would also announce pre-release versions.

github.cake
augustoproiete: > If we add an `addin.cake` it just needs to be loaded once in build.cake

I understand that this is all that is needed for the build process to work due to the concatenation of files, etc. but if one opens `github.cake` on VS Code, etc. would they get proper intellisense without declaring the addins that this file relies on? (I would think not, which is why I suggested a shared `addins.cake` would be added to all files instead of only `build.cake`)

I did a [quick test](https://github.com/augustoproiete-sandbox/test-multiple-cake-includes) and Cake is smart enough to #load a file only once, so that seems viable (?)

github.cake
augustoproiete: I'm happy to refactor into a build task with shared context as that's not something I've used before, so a good excuse to try it out.

The `_cakeGitHubReleaseInfo` is not used anywhere but in the extension method so whilst technically a global variable, it's not accessed by any other part of the build - that's what I meant.

github.cake
pascalberger: > but if one opens github.cake on VS Code, etc. would they get proper intellisense without declaring the addins that this file relies on?

Yes, this should work following [Enabling IntelliSense](https://cakebuild.net/docs/integrations/editors/vscode/intellisense#enabling-intellisense), especially that the folder needs to be opened in VSCode and not a single file. In this case all Cake files will be concatenated and passed to OmniSharp as one big string. There are also lot of other reasons why IntelliSense for a single file won't work (e.g. global variables).

github.cake
augustoproiete: Got it. It seems that the end goal is that the `Preview` task would not access any external resource and just build using local resources. So in preview we would have to either use a fallback value for these variables or display empty (e.g. in the [getting started markdown page](https://github.com/cake-build/website/pull/1462/commits/f428886e85c628d50f11d07ffd78f693138baee5#diff-4e129ca53399e95e0d8764573e9142df4c72e3c7cd70a7d4ddda1cb565c8e7df))

github.cake
pascalberger: I still think we can merge this first and do the refactoring later. Also to keep this PR simpler and more focused. Introduction of a typed context would touch a lot of stuff in the build script not related to showing the version on home page. And from a functional point of view the implementation works fine.

github.cake
augustoproiete: I understand the concat process, and the scoping. What I tried to convey is:

1. It's not very clear which scripts are meant to be executed directly (e.g. `build.cake`, `deploy.cake`) and which scripts are meant to be loaded by other scripts only (`nuget.cake` and now `github.cake`) so perhaps we could use different file names to make the intent (and maybe the relationship) more explicit

2. If one of those separate files (e.g. `github.cake`) would use a variable that is not declared within the file, it's not very clear where the variable is declared (could be `build.cake`, `nuget.cake`, ...) If `deploy.cake` needed some `github.cake` function, now it would also have to declare the same global variables that `github.cake` expects it to have, and declare it the same way

Of course, I'm looking at the above in the shoes of new contributors... I'm familiar enough now with the website build scripts to know where things are and what the execution chain is.

pascalberger: 1: The way I've solved this and seen in other repos is with a  `build.cake` file in the root folder and other Cake files loaded by the `build.cake` file in a `build` subfolder. 
2: The thing is it doesn't matter where global variables are declared. If a variable is declared in `build.cake` and used in `nuget.cake`, then it will also work if the method in `nuget.cake` is called from `deploy.cake`. I agree with the point that code can be hard to read with global variables. Writing helper methods in a way that everything is passed in as arguments would get rid of the dependencies. It's actually like with any .NET program :)

pascalberger (approved):
I removed the additional Cake.Yaml declaration in the `github.cake` file so that we can merge this. Additional refactoring can be done in subsequent PRs

pascalberger:

PR #1488, opened 2021-01-19
Remove Cake specific addin categories

pascalberger:
Remove addin categories which identify addins as Cake addins since this doesn't make sense when shown on the Cake website.


gep13 (approved):
LGTM!

PR #1506, opened 2021-01-24
Use "Cake Frosting" instead of "Cake.Frosting" everywhere

pascalberger:
Uses everywhere `Cake Frosting` when mentioning Frosting, instead of `Cake.Frosting`


augustoproiete (approved):
LGTM

PR #1519, opened 2021-01-30
(build) Update to use lightweight bootstrapper with dotnet tool

augustoproiete:
From https://github.com/cake-build/resources/tree/master/dotnet-tool


build.cmd
```diff
@@ -1,2 +1,11 @@
-@echo off
-powershell -ExecutionPolicy Unrestricted ./build.ps1 %CAKE_ARGS% %*
\ No newline at end of file
+@echo on
```
pascalberger, line 1: What's the reason to have a build.cmd boostrapper when we already have a build.ps1? Previously it at least called internally the Powershell bootstrapper, but with this change we will have duplicated logic.

augustoproiete: The `.cmd` itself is for convenience for devs that don't use PowerShell and instead use Command prompt/Windows terminal (cmd), so I think it make sense to have it (regardless if it just calls the `.ps1` or not).

Now about having the duplicate commands, it's really personal preference. If the .ps1 bootstrapper was complex, calling the `.ps1` would be the only sane choice, but nowadays the required code in the `.cmd` is so simple now, that I prefer to have all the commands there and know it's going to work for sure, regardless of the state or version of PowerShell I have on the machine.

No strong feelings on this one though and happy to call PS if the duplicate commands triggers some devs OCD's :smile:

PR #1521, opened 2021-02-01
Update working directory handling in Frosting Getting Started Guide

pascalberger:
`UseWorkingDirectory` can lead to some known issues (e.g. with tool installation). We should therefore not promote its usage in the getting started guide.


devlead (approved):
LGTM 👍

PR #1525, opened 2021-02-02
Add a point to addin guidelines that addins should be tested with different runners

pascalberger:
Add a point to addin guidelines that addins should be tested with different runners


augustoproiete (approved):
LGTM

nils-a (approved):
LGTM

PR #1589, opened 2021-02-09
Add example how to pass variables from CI systems

pascalberger:
Add example how to pass variables from CI systems to CLI with 1.0


nils-a (approved):
LGTM

PR #1598, opened 2021-02-11
Increase number of search results to 10

pascalberger:
For some searches the current shown 5 search results can be a lit bit limiting. This PR increase number of search results shown to 10.


augustoproiete (approved):
LGTM!

PR #1599, opened 2021-02-11
Add search box for searching DSL section

pascalberger:
Adds a search box to the DSL reference main page which searches only through all pages below `/dsl`.

image


augustoproiete (approved):
LGTM! Great idea using Algolia search with a narrower scope :shipit:

PR #1622, opened 2021-02-14
Update targets documentation for Frosting

pascalberger:
Update documentation how to run targets for Frosting

### DIFF
--- a/input/docs/writing-builds/running-targets.md
+++ b/input/docs/writing-builds/running-targets.md
@@ -3,7 +3,10 @@ Description: How to run targets and tasks
 RedirectFrom: docs/fundamentals/running-targets
 ---
 
-To run a target, use the `RunTarget` method. The `RunTarget` method should be placed at the end of the script.
+# Cake script runner
+
+When using [Cake .NET Tool], [Cake runner for .NET Framework] or [Cake runner for .NET Core] you can use the `RunTarget` method to run a target.
+The `RunTarget` method should be placed at the end of the script.
 
 ```csharp
 Task("Default")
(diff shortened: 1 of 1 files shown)


nils-a:
@pascalberger from the text I inferr that there is currently no way in code to use another task as default than `Default`. Should we make that clearer?

pascalberger:
> @pascalberger from the text I inferr that there is currently no way in code to use another task as default than `Default`. Should we make that clearer?

Yes, currently you can't define task in code. I tried to make it clear with the following sentence

> @pascalberger from the text I inferr that there is currently no way in code to use another task as default than `Default`. Should we make that clearer?

Yes this is currently not possible. I tried to make it clear with the `Cake Frosting by default will call a task named Default.` sentence. Now added an additional annotation. 

@nils-a Please let me know if you have better suggestions how it would be more clear for you.

input/docs/writing-builds/running-targets.md
```diff
@@ -49,3 +52,31 @@ The following arguments will run the `Publish` target without running the `Build
 ```powershell
 --target=Publish --exclusive
 ```
+
+# Cake Frosting
+
+[Cake Frosting] by default will call a task named `Default`.
+
+:::{.alert .alert-info}
+In [Cake Frosting] it is not possible in to define the task which is run by default in your build script.
```
nils-a, line 61:
```suggestion
In [Cake Frosting] it is not possible in to define a task other than `Default`, which is run by default in your build script.
```

nils-a (approved):
One minor change, but not crucial. 
LGTM

PR #1648, opened 2021-02-19
Update Pinning Cake Version examples to use Cake v1.0.0

augustoproiete:
...

### DIFF
--- a/input/docs/writing-builds/reproducible-builds/pinning-cake-version.md
+++ b/input/docs/writing-builds/reproducible-builds/pinning-cake-version.md
@@ -20,7 +20,7 @@ To have deterministic builds it is important that on every build the same versio
     <div id="frosting1" class="tab-pane fade">
         <p>
             Make sure to use a fixed version number for <code>Cake.Frosting</code> in the <code>*.csproj</code> file:<br/>
-<pre><code class="language-xml hljs">&lt;PackageReference Include="Cake.Frosting" Version="0.38.4" /&gt;</code></pre>
+<pre><code class="language-xml hljs">&lt;PackageReference Include="Cake.Frosting" Version="<?! Meta CakeLatestReleaseName /?>" /&gt;</code></pre>
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


nils-a:
Should we add, for `Cake .NET Tool`, that the manifest should be placed under version control?

input/docs/writing-builds/reproducible-builds/pinning-cake-version.md
```diff
@@ -20,7 +20,7 @@ To have deterministic builds it is important that on every build the same versio
     <div id="frosting1" class="tab-pane fade">
         <p>
             Make sure to use a fixed version number for <code>Cake.Frosting</code> in the <code>*.csproj</code> file:<br/>
-<pre><code class="language-xml hljs">&lt;PackageReference Include="Cake.Frosting" Version="0.38.4" /&gt;</code></pre>
+<pre><code class="language-xml hljs">&lt;PackageReference Include="Cake.Frosting" Version="1.0.0" /&gt;</code></pre>
```
pascalberger, line 23: In the meantime we know what the latest version is (same functionality which shows it on home page). Instead of using a fixed version we could use the dynamic latest version

augustoproiete: I'm in favor of this - I was not sure if the fact that Frosting is currently documented as 0.38.4 and Scripting as 0.38.5 was on purpose or not

augustoproiete:
> Should we add, for `Cake .NET Tool`, that the manifest should be placed under version control?

@nils-a I think it's a good idea. I'll add this in.

input/docs/writing-builds/reproducible-builds/pinning-cake-version.md
augustoproiete: :white_check_mark: Updated to use latest version of Cake dynamically

augustoproiete:
@nils-a I've added the manifest/source control wording.

nils-a (approved):
LGTM!

PR #1703, opened 2021-02-25
(GH-1702) Improve documentation for empty switches

### LINKED ISSUE #1702: Improve documentation for passing empty values for an argument
To pass empty values for arguments to Cake 1.0 a space needs to be used instead of `=`. We already mention it in [upgrade instructions](https://cakebuild.net/docs/getting-started/upgrade#passing-empty-arguments), but could make it more clear that it also applies for empty values (`--foo=""`).

Additionally we should document this on the runner page.

See also https://github.com/cake-build/cake/discussions/3198

pascalberger:
Improve documentation for switches without or with and empty value.

Fixes #1702

### DIFF
--- a/input/Shared/switches.txt
+++ b/input/Shared/switches.txt
@@ -32,3 +32,11 @@
 
 All switches not recognized by Cake will be added to an argument list that is passed to the build script.
 See [Arguments And Environment Variables](/docs/writing-builds/args-and-environment-vars#arguments) how to read arguments in your script.
+
+:::{.alert .alert-info}
+Use a space instead of `=` if the value of a switch can be empty or null:
+
+```
+--foo ""
+```
+:::
--- a/input/docs/getting-started/upgrade.md
+++ b/input/docs/getting-started/upgrade.md
@@ -162,7 +162,7 @@ Error: Unknown command 'Foo'.
 
 ### Passing empty arguments
 
-With previous versions of Cake it was possible to define an empty argument (e.g. `--foo=`).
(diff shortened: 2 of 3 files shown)


input/Shared/switches.txt
```diff
@@ -32,3 +32,11 @@
 
 All switches not recognized by Cake will be added to an argument list that is passed to the build script.
 See [Arguments And Environment Variables](/docs/writing-builds/args-and-environment-vars#arguments) how to read arguments in your script.
+
+:::{.alert .alert-info}
+Use a space instead of `=` if the value of a switch can be empty or null:
```
nils-a, line 37: Is "empty or null" precice enough? My tests always result in `null`, regardless whether I set `--foo` or `--foo ""`.

augustoproiete (approved):
LGTM

input/Shared/switches.txt
augustoproiete: I think it's clear, and the code example after that helps. How would phrase it differently?

Also, what do you mean by "_my tests always result in `null`_"

In my tests, both `--foo` and `--foo ""` results in the argument not being present in the collection of arguments, which means that `Argument<string>("foo")` throws an exception on both cases with `Argument 'foo' was not set`.

So the only way to get `null` is to specify `null` as the default value i.e. `Argument<string>("foo", null)` but in that case you're only getting `null` because you said that was the default... Not because Cake decided for you. The same way that `Argument<string>("foo", "abc")` will give you `abc` for both `--foo` and `--foo ""`

nils-a (approved):
LGTM

PR #1738, opened 2021-03-06
1.1.0 blog post

augustoproiete:
...

### DIFF
--- a/input/blog/2021-03-06-cake-v1.1.0-released.md
+++ b/input/blog/2021-03-06-cake-v1.1.0-released.md
@@ -1,0 +1,100 @@
+﻿---
+title: Cake v1.1.0 released
+category: Release Notes
+author: augustoproiete
+releaseName: 1.1.0
+---
+
+Version 1.1.0 of Cake has been released. Take it for a spin and give us feedback on our [discussion board](https://github.com/cake-build/cake/discussions).
+
+This is the first release since our big [Cake 1.0 milestone release](/blog/2021/02/cake-v1.0.0-released), and includes new features, improvements and bug fixes to both [Cake Scripting](/docs/running-builds/runners/dotnet-tool) and [Cake Frosting](/docs/running-builds/runners/cake-frosting).
+
+### Highlights of this release
+
+- .NET tools are now natively supported in Cake builds via `#tool dotnet:`
+  - The [Cake.DotNetTool.Module](https://github.com/cake-contrib/Cake.DotNetTool.Module) is bundled with Cake from now on and is no longer required as a separate package/dependency
+- Cake Frosting: Tools are now installed in the working directory by default, and Frosting now honors the `Tools` path defined in `cake.config` the same way that Cake scripting does
+- TeamCity: Additional properties of TeamCity builds is now available to Cake builds
+- Azure Pipelines: Output variables in Azure Pipelines can be used to share build data between different stages and/or different tasks
+- Improved tool resolution to prefer tools compatible with the platform where the build is running on
+- Several improvements on coloring of console output when running Cake builds:
+  - Colored build output is now enabled by default when running Cake on Azure Pipelines and TeamCity
+  - We removed the default black background on build log messages and we now use the default background color of the console, which makes build logs look great if you use a custom background color, such as PowerShell
+  - It's now possible to turn off coloring of console output by defining an environment variable called [`NO_COLOR`](https://no-color.org)
+
+... and more! The full list of changes is [listed below](#issues).
+
+### Contributors
+
+This release was made possible thanks to the contribution of these awesome members of the [Cake community](/community/thanks/) along with [Cake team](/docs/team/):
+
+- [augustoproiete](https://github.com/augustoproiete)
+- [gep13](https://github.com/gep13)
+- [patriksvensson](https://github.com/patriksvensson)
+- [devlead](https://github.com/devlead)
+- [FroggieFrog](https://github.com/FroggieFrog)
+- [Marusyk](https://github.com/Marusyk)
+- [gitfool](https://github.com/gitfool)
+- [Cheesebaron](https://github.com/Cheesebaron)
+- [soroshsabz](https://github.com/soroshsabz)
+- [nmbro](https://github.com/nmbro)
+- [BlythMeister](https://github.com/BlythMeister)
+- [Jericho](https://github.com/Jericho)
+- [prochnowc](https://github.com/prochnowc)
+
+Full details of everything that is included in this release can be seen [below](#issues).
+
+### Upgrading
+
+See [Upgrade instructions](/docs/getting-started/upgrade) for documentation about how to upgrade to a newer Cake version.
+
+<!--excerpt-->
+
+### Issues
+
+As part of this release we had [33 issues](https://github.com/cake-build/cake/milestone/77?closed=1) closed.
+
+__Feature__
+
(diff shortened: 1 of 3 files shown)


pascalberger (changes requested):
Two smaller issues

devlead (approved):
Beyond @pascalberger feedback and upstream rebase needed looks good to me 👍

input/blog/2021-03-06-cake-v1.1.0-released.md
```diff
@@ -0,0 +1,120 @@
...
+
+Version 1.1.0 of Cake has been released. Take it for a spin and give us feedback on our [discussion board](https://github.com/cake-build/cake/discussions).
+
+This is the first release since our big [Cake 1.0 milestone release](/blog/2021/02/cake-v1.0.0-released), and includes new features, improvements and bug fixes to both [Cake Scripting](/docs/running-builds/runners/dotnet-tool) and [Cake Frosting](/docs/running-builds/runners/cake-frosting).
+
+### Highlights of this release
+
+- .NET tools are now natively supported in Cake builds via `#tool dotnet:`
+  - The [Cake.DotNetTool.Module](https://github.com/cake-contrib/Cake.DotNetTool.Module) is bundled with Cake from now on and is no longer required as a separate package/dependency
+- Cake Frosting: Tools are now installed in the working directory by default, and it's Frosting now honors the `Tools` path defined in a `cake.config` the same way that Cake scripting does
```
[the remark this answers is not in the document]
augustoproiete: :white_check_mark:

input/blog/2021-03-06-cake-v1.1.0-released.md
```diff
@@ -0,0 +1,120 @@
...
+# If you are using the .NET tool as a local tool for the repository
+dotnet tool install --local Cake.Tool
+
+# If you are using the .NET tool as a global tool for the user
+dotnet tool install --global Cake.Tool
+```
+
+If you are upgrading from Cake 0.38.x, see [Upgrade instructions](/docs/getting-started/upgrade#cake-0.38.x-to-cake-1.0) for documentation about breaking changes in this release.
+
+#### Note about Cake.DotNetTool.Module
```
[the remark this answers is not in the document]
augustoproiete: :white_check_mark:

input/docs/getting-started/upgrade.md
```diff
@@ -3,6 +3,28 @@ Title: Upgrade instructions
 Description: Upgrade instructions between different versions of Cake
 ---
```
pascalberger, line 5: Add heading: `Cake 1.0 to Cake 1.1`

input/docs/getting-started/upgrade.md
```diff
@@ -3,6 +3,28 @@ Title: Upgrade instructions
...
+# If you are using the .NET tool as a local tool for the repository
+dotnet tool install --local Cake.Tool
+
+# or
+
+# If you are using the .NET tool as a global tool for the user
+dotnet tool install --global Cake.Tool
+```
+
+### Note about Cake.DotNetTool.Module
```
pascalberger, line 18: Focus the heading to the essential part:

```suggestion
Cake.DotNetTool.Module
```

input/docs/getting-started/upgrade.md
```diff
@@ -3,6 +3,28 @@ Title: Upgrade instructions
...
+
+If you use `Cake.DotNetTool.Module` already on your builds, you should either:
```
pascalberger, line 20: An unordered list with a number first looks strange. As it is an either/or choice an unordered list should work fine:

```
If you use `Cake.DotNetTool.Module` already on your builds do one of the following:

- Upgrade Cake to version 1.1.0, and _remove_ the `Cake.DotNetTool.Module` from your build script (as it's no longer needed).
- Pin `Cake.DotNetTool.Module` to a version compatible with the Cake version that you use. `Cake.DotNetTool.Module` version 1.0.1 is the last version compatible with Cake 1.0.0. Starting with this release, `Cake.DotNetTool.Module` will always be released together with Cake and will only be compatible with the current release.
```

pascalberger (changes requested):

input/docs/getting-started/upgrade.md
augustoproiete: I thought about this, but that means we need to update this heading on every release... Do we want that :thinking:?

augustoproiete: I gave it a go with the variable version - the downside is that it's not going to generate  stable URL (i.e. changes on every release). Given it's at the beginning of the page, doesn't seem like a big issue

pascalberger: Yes, that was the overall idea of the page to have a section with instructions how to update from one release to another. What's the issue with that? I'm not speaking about the page title (which is `Upgrade instructions`). I'm speaking about adding an additional `# Cake 1.0 to Cake 1.1` heading before the content relevant for updating from 1.0 to 1.1.

input/docs/getting-started/upgrade.md
augustoproiete: :white_check_mark:

input/docs/getting-started/upgrade.md
pascalberger: > I gave it a go with the variable version - the downside is that it's not going to generate a stable URL (i.e. changes on every release). Given it's at the beginning of the page, doesn't seem like a big issue

We speak about different things. There should be nothing dynamic on this page. The intent was to have a page with a section containing upgrade instructions for every version.

```
# Cake 1.0 to Cake 1.1

Blablabla

# Cake 0.38.x to Cake 1.0

Blablabla
```

augustoproiete: The thing is that the upgrade instructions for Cake 1.1.0 to Cake 1.2.0 and 1.3.0 and probably 2.0.0 will be the same: `dotnet tool update Cake.Tool`, etc.

That's why I didn't put a header because this is evergreen until .NET tool is replaced by something else

input/docs/getting-started/upgrade.md
augustoproiete: :white_check_mark:

input/docs/getting-started/upgrade.md
pascalberger: Like it is now doesn't make much sense. `Cake.DotNetTool.Module` is specific to this version and should be below a `Cake 1.0 to Cake 1.1` heading. If we want to have general tool instructions here we should provide them for all supported runners and not only for one (like e.g. done here: https://cakebuild.net/docs/writing-builds/reproducible-builds/pinning-cake-version). But I would suggest to do this in a separate PR to not block the 1.1 blog post

augustoproiete: :thinking: I'm not sure I agree that `Cake.DotNetTool.Module` (from the point of view of upgrade instructions) is specific to 1.1, because not all users will be upgrading from 1.0 to 1.1... Many will be upgrading from and old 0.38.x, skipping a few 0.38.x versions and 1.0, and going straight into 1.1.

Anyone using the `Cake.DotNetTool.Module` in any old version of Cake, would want to read this call-to-action... To me it's a generic piece of upgrade instructions that should stay there for a while and kept visible when 1.2 comes out, etc.

But anyway, I moved the header down and put the generic instructions above the header, so seems like a good compromise and we can decide what to do in 1.2 or 2.0 whichever comes first.

image

pascalberger (approved):

Issue #1739, opened 2021-03-06
Document changes in DotNetTool Module starting with Cake 1.1.0


augustoproiete:
### Recommendation

- If you use `Cake.DotNetTool.Module` already on your builds, you should either:

- (1) Upgrade Cake to version 1.1.0, and _remove_ the `Cake.DotNetTool.Module` from your build script (as it's no longer needed)

or

- (2) Pin `Cake.DotNetTool.Module` to a version compatible with the Cake version that you use. `Cake.DotNetTool.Module` version 1.0.1 is the last version compatible with Cake 1.0.0. Starting with this release, `Cake.DotNetTool.Module` will always be released together with Cake and will only be compatible with the current release.

TomerPacific:
Hi @augustoproiete, I'd like to help out. What are the changes you would like done exactly?

augustoproiete:
Hi @TomerPacific thanks for your interest in this one.

The intent is to add the recommendation above to this page here: https://cakebuild.net/extensions/cake-dotnettool-module/

TomerPacific:
@augustoproiete, I've taken a look inside the project to see how I can add what you have written and am coming up short.
I found inside **extensions** folder the file for the DotNetToolModule, but that is a yml file and I don't think I am supposed to add that there (based on looking at other extensions as well).

I've been looking at the [7zip extension](https://cakebuild.net/extensions/cake-7zip/) as a go to guide to follow, but have yet to find out how the Aliases text has been added there. From a brief search I found out the **config.wyam** file, but I am not sure how it all fits in together. I would appreciate it if you could declutter some of this.

augustoproiete:
@TomerPacific Thanks for looking deeper into this one.

We [discussed](https://github.com/cake-build/website/pull/1743) this a while ago, and I believe the easiest path forward is to add a hard-coded alert in the page for now (e.g. `if (extensionName == "Cake.DotNetTool.Module") @Html.Partial("_DotNetToolModuleDescription"); ...` given that we don't yet have a feature that allows us to add custom notes about a specific extension.

![image](https://user-images.githubusercontent.com/177608/135932961-005353d5-390c-48c4-95f7-e4161e7fc926.png)

As part of this change, we should also rollback the last changes that were made to `AnalyzedPackageVersion` and `AnalyzedPackagePublishDate` to reflect the latest "valid" version of this extension, as per @pascalberger's [comments here](https://github.com/cake-build/website/pull/1743#issuecomment-792333452):

```diff
[...]

PR #1743, opened 2021-03-07
Remove Cake.DotNetTool.Module as it ships now with Cake

pascalberger:
Remove Cake.DotNetTool.Module as it ships now with Cake since 1.1 (see https://github.com/cake-build/cake/issues/2903). 

Related PR to exclude in addin discoverer: https://github.com/cake-contrib/Cake.AddinDiscoverer/pull/184


augustoproiete:
LGTM.

However, I'm wondering :thinking: if it wouldn't be better to leave this module in the website for the time being, and add an alert on its page with the same text that we have on the blog post, e.g.:

image

It could be hard-coded in the page for now (e.g. `if (extensionName == "Cake.DotNetTool.Module") ...` just to have something quick, given that we don't yet have a feature to add custom notes about a specific extension.

pascalberger:
Thought about this as well. It was only added recently and there and we generally only document current version on website and not previous versions (e.g. removed command line switches). 

On the other hand it is not a big deal to add a not as you suggested and would be OK for me. To be consistent we should in this case document the latest version of the extension which was not bundled (1.0.1).

@Jericho If we keep the YAML file of the old version here, would it work with excluding it in Addin Discoverer? Or will it then try to remove it?

Jericho:
> @Jericho If we keep the YAML file of the old version here, would it work with excluding it in Addin Discoverer? Or will it then try to remove it?

Excluding this module means that discoverer will neither update nor delete this yaml which, if I understand correctly, is exactly what you want.

Jericho:
Discoverer 3.70.0 published. The module is now excluded from analysis.

augustoproiete:
Closing this one now that `Cake.DotNetTool.Module` has been blacklisted in AddinDisco. We'll document it via https://github.com/cake-build/website/issues/1739

PR #1746, opened 2021-03-07
Mark modules containing tool providers

pascalberger:
Mark modules containing tool providers which allows to search for them


augustoproiete (approved):
LGTM

PR #1747, opened 2021-03-07
Split preprocessor documentation into separate pages

pascalberger:
Split preprocessor documentation into separate pages in preparation to update and improve the content (e.g. for .NET Tool support). 

This PR intentionally does not update the content to make reviewing easier. Content will be updated in subsequent PRs.

The part about the [Static usings](https://cakebuild.net/docs/writing-builds/preprocessor-directives#using-static-directive) has been removed though, since this is actually not a preprocessor directive. If we still want to have this documented we better move it to a Tips and Tricks section.


augustoproiete (approved):
LGTM.

It looks like the URL stays the same i.e. `preprocessor-directives.md` -> `index.cshtml`.

Is that true? If not, I'd suggest adding a redirect.

pascalberger:
> It looks like the URL stays the same i.e. `preprocessor-directives.md` -> `index.cshtml`.

https://cakebuild.net/docs/writing-builds/preprocessor-directives will still be available afterwards. In theory https://cakebuild.net/docs/writing-builds/preprocessor-directives.html is also available currently, which won't be afterwards, but this is something which we didn't link from anywhere.

PR #1765, opened 2021-03-09
Mention the default target for Cake.Frosting

mgnslndh:
I did not find any information about the default target that would run when not using the `--target` command line switch. Should it be improved or will this work?


pascalberger (approved):

augustoproiete:
@mgnslndh your changes have been merged, thanks for your contribution :+1:

PR #1992, opened 2021-05-08
Use updated cake-contrib icons for extensions

pascalberger:
Use updated cake-contrib icons for extension listings


augustoproiete (approved):
Nice!

PR #2225, opened 2021-10-11
(GH-1248) Add toggle switch to show only built in aliases

### LINKED ISSUE #1248: Add toggle switch to show only built in aliases
Current reference page shows all categories with built-in aliases and aliases from addins. For users not interested in using addins it might be useful to have a toggle switch on that page to filter out categories which only contain aliases from addins, and also filter out aliases from addins in the category detail page.

pascalberger:
Add toggle switch to alias reference page to show only built in aliases.

image

* If toggle is off, all categories containing only aliases from addins will be hidden from left sidebar
* If toggle is off, aliases from addins no longer will be listed for categories containing built-in aliases and aliases from addins
  * There's currently a small issue that they still are shown in the TOC of the page
* Toggle state is persisted across page calls in browser storage
* On pages containing only aliases from addins toggle is disabled, since otherwise page will become empty and be hidden from navigation when toggle is switched off
* On pages containing only aliases form addins persisted browser storage is ignored and toggle is always enabled (use-case is if user has toggle switched off, but accesses page through a direct link)
* Default value of toggle is `On`

Fixes #1248

### DIFF
--- a/config.wyam
+++ b/config.wyam
@@ -57,6 +57,13 @@ Pipelines.InsertAfter("Aliases", "DslAliases",
             .Distinct(),
         Documents("Aliases")
     ),
+    Meta(
+        "ContainsBuiltInAliases",
+        @doc
+            .DocumentList(Keys.GroupDocuments)
+            .Any(x => string.IsNullOrWhiteSpace(x.Document(CodeAnalysisKeys.ContainingAssembly)?.String(CodeAnalysisKeys.DisplayName)))
+            ? true
+            : false),
     Meta(Keys.WritePath, new FilePath("dsl/" + @doc.String(Keys.GroupKey).ToLower().Replace(" ", "-") + "/index.html")),
     Meta(Keys.RelativeFilePath, @doc.FilePath(Keys.WritePath)),
     OrderBy(@doc.String(Keys.GroupKey))
--- a/input/_DslLayout.cshtml
+++ b/input/_DslLayout.cshtml
@@ -47,35 +47,45 @@
             }
         }
 
-        // Group heading
-        var groupHeading = addinName ?? "Built-In";
-        <h1 id='@groupHeading.Replace(" ", "-")'>@groupHeading</h1>
+        var aliasClass = isAddin ? "addinAliases" : "builtinAliases";
+        var aliasStyle = isAddin ? "style=display:none;" : null;
+        <div class="aliasList @aliasClass" @aliasStyle>
+            @{
+                // Group heading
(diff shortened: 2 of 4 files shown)


pascalberger:
> Is it worth considering a static generation approach where cake-only aliases get generated a second time on a separat folder/URL, and we resolve the toggle without the need of client-side scripting?

Server side generated pages is what we had until a year ago. If we want to go back to this then we should just revert https://github.com/cake-build/website/pull/1247.

But this approach was far from perfect and had a lot of issues, like that for example built-in and addins can share categories. See https://github.com/cake-build/website/issues/1078 for details.

List.js is also the library which we already use for extension search, which is also client side. If we want to go server-side we should change to server-side there to. But on my side I don't see any performance issues, nor do I see how user experience should be better with server-side rendered pages, which would require a full page reload every time some filtering is changed. We will also lose persistence of user preference (except we still add some client side code together with additional routing). All in all I don't see any advantages in implementing this server-side, on the contrary even more effort to implement, and don't see anybody picking up stuff like this (e.g. this issue was open for a year).

> On a separate note, (_understandably_) the search is not affected by the toggle, which might be confusing, so maybe for now we should add an alert that explains the search is across all aliases, if the toggle is `off`?

Sure, it is not perfect and finished, but one more step in the direction what was started with https://github.com/cake-build/website/issues/1078. Search experience was something which I wanted to look after this PR. Agree that it either should take toggle into account or explain how it works. There are also other improvements which can be made on top of this PR like, https://github.com/cake-build/website/pull/2228 to introduce alias category filtering.

> I wonder if the static generation approach I mention above wouldn't also be something useful to enable us getting a filtered search across cake-only aliases 🤔

Having built-in aliases and aliases for addins under different folders would make it easier to generate separate indexes in Doc Search. But I don't think it is worth it, considering the disadvantages of the approach as explained above. There might also be other possibilities with current structure to get to the same result.

pascalberger:
@augustoproiete Can you please give it another spin. I made some improvements for rendering of toggle and sidebar navigation which might solve the issues you saw before.

augustoproiete:
> @augustoproiete Can you please give it another spin. I made some improvements for rendering of toggle and sidebar navigation which might solve the issues you saw before.

(_I tested latest as-of `a84edd4` rebased on top of develop which produced `c458248`_)

The flickering got better after the latest visibility changes, but still somewhat noticeable - if it doesn't bother anyone else, happy to move this along.

With the toggle (this PR):

after

Before the toggle (`master`):

before

PR #2305, opened 2021-12-02
Use Cake.Core to determine available versions of Cake

pascalberger:
Use `Cake.Core` to determine available versions of Cake, since `Cake` is not longer updated since 2.0


nils-a (approved):
LGTM

PR #2408, opened 2022-03-15
Update videos page with YouTube link

corbob:
The Channel 9 video is no longer at the previous URL. It appears to be available on YouTube, so using the link to that video.

Note: This was the video that YouTube found when I went searching for it. I am unsure if the original was moved elsewhere by Microsoft, or if they have a better host for it.


devlead:
@corbob your changes have been merged, thanks for your contribution 👍

PR #2518, opened 2022-11-08
Add information about new setting added in v3.0.0

gep13:
This adds information about the setting that was added in this PR: https://github.com/cake-build/cake/pull/4056

### DIFF
--- a/input/docs/running-builds/configuration/default-configuration-values.md
+++ b/input/docs/running-builds/configuration/default-configuration-values.md
@@ -81,6 +81,100 @@ Addins=./tools/Addins</code></pre>
     </div>
 </div>
 
+#  Cache compiled script on disk
+
+:::{.alert .alert-info}
+Available since Cake `2.2.0`.
+:::
+
+This enables [Cake .NET Tool] to cache script compilation for significantly faster subsequent executions of unchanged scripts.
+
+_Default value:_ `false`
+
+_Valid values:_ `true`, `false`
+
+<ul class="nav nav-tabs">
+    <li class="active"><a data-toggle="tab" href="#env11">Environment variable name</a></li>
+    <li><a data-toggle="tab" href="#config11">Config file contents</a></li>
+    <li><a data-toggle="tab" href="#arg11">Argument name</a></li>
+</ul>
+
+<div class="tab-content">
+    <div id="env11" class="tab-pane fade in active">
+        <p>
+            <pre><code class="language-sh hljs">CAKE_SETTINGS_ENABLESCRIPTCACHE</code></pre>
+        </p>
+    </div>
+    <div id="config11" class="tab-pane fade">
+        <p>
+            <pre><code class="language-sh hljs">[Settings]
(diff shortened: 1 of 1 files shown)


gep13:
@devlead @pascalberger I changed the order of some of the configuration options, to make them alphabetical, is there anything "extra" I need to do with the `arg`, `environment`, and `config` attributes? Or will these continue to work, even though I have moved the sections to a different location in the page?

pascalberger:
> @devlead @pascalberger I changed the order of some of the configuration options, to make them alphabetical, is there anything "extra" I need to do with the `arg`, `environment`, and `config` attributes? Or will these continue to work, even though I have moved the sections to a different location in the page?

Location doesn't matter, as long as every one has a unique id. Otherwise tabs will switch together.

gep13:
@pascalberger sweet!  The new section I added has `arg13`, etc, which I saw as the next available number.

input/docs/running-builds/configuration/default-configuration-values.md
```diff
@@ -342,96 +436,39 @@ Tools=./tools</code></pre>
...
-<div class="tab-content">
-    <div id="env10" class="tab-pane fade in active">
-        <p>
-            <pre><code class="language-sh hljs">NO_COLOR</code></pre>
-        </p>
-    </div>
-</div>
-
-#  Cache compiled script on disk
+# Use Spectre Console For Console Output
```
pascalberger, line 439: Is `Spectre Console` a term we can expect users to be aware of? Otherwise I would suggest to call it `advanced console output`

```suggestion
# Enable advanced console output
```

input/docs/running-builds/configuration/default-configuration-values.md
```diff
@@ -342,96 +436,39 @@ Tools=./tools</code></pre>
...
 
 :::{.alert .alert-info}
-Available since Cake `2.2.0`.
+Available since Cake `3.0.0`.
 :::
 
-This enables [Cake .NET Tool] to cache script compilation for significantly faster subsequent executions of unchanged scripts.
+When enabled (i.e. set to `true`), the console output from Cake is rendered via Spectre.Console. This includes an additional column in the Task Summary table at the end of a build, which lists the reason (if available), on why a task was skipped during the build.
```
pascalberger, line 445: We can make this a link

```suggestion
When enabled (i.e. set to `true`), the console output from Cake is rendered via [Spectre.Console](https://spectreconsole.net/) library. This includes an additional column in the Task Summary table at the end of a build, which lists the reason (if available), on why a task was skipped during the build.
```

pascalberger (approved):
LGTM. Two suggestions

pascalberger (changes requested):
The PR contains a `theme` submodule which is not required

gep13:
@pascalberger I have no idea where that came from, but I have fixed it. Thanks for catching that!

pascalberger (approved):
LGTM

PR #2544, opened 2023-01-18
Add Cake.AzureCli

lprichar:


lprichar:
I just realized this probably won't ever get merged because it's basically a duplicate of https://github.com/cake-build/website/pull/888.  Should I just close it?

pascalberger:
> I just realized this probably won't ever get merged because it's basically a duplicate of #888. Should I just close it?

I already wanted to mention that we currently don't have it added on purpose. 

What confused me, was that when looking at the build time of this PR, build time didn't increase. But hadn't time to investigate this yet, and if the build even builds the API for Cake.AzureCli.

PR #2619, opened 2023-06-12
(maint) Update to latest Google Analytics

gep13:
The Google Analytics property that we currently have is set to be retired on 1 July 2023.  If you log in, it states:

This property will stop processing data starting 1 July 2023. Use the GA4 Setup Assistant to complete the next steps.

A new property has been created, and this commit updates to use this new property.

### DIFF
--- a/input/_Bottom.cshtml
+++ b/input/_Bottom.cshtml
@@ -68,14 +68,12 @@
     }
 </script>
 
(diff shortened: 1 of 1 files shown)


gep13:
Going to close this PR since we are removing the usage of Google Analytics.

PR #2677, opened 2023-07-23
Use list of contributors as provided by Addin Discoverer

### LINKED ISSUE #2243: Rate limiting contributor list
[Contributor list](https://cakebuild.net/community/thanks/#contributors) is currently retrieved client side where we sometimes run into rate limiting issues.

We could make authenticated calls to get a higher rate limit or create the list server side

pascalberger:
Create list of contributors from the static YAML file maintained by addin discoverer instead of reading them on every client request. This helps to ensure that we don't run into rate limiting when reading contributors.

Fixes #2243

### DIFF
--- a/config.wyam
+++ b/config.wyam
@@ -21,6 +21,12 @@ Pipelines.InsertBefore(Docs.Code, "Maintainers",
     Yaml()
 );
 
+// Reads contributor metadata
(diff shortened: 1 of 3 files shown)


augustoproiete (approved):
LGTM

PR #2679, opened 2023-07-23
Remove unnecessary parsing of maintainer files in build script

pascalberger:
Remove unnecessary parsing of maintainer files in build script, since the actual parsing is done by Wyam


augustoproiete (approved):
LGTM

PR #2703, opened 2023-08-12
Return cake-contrib commits on contributors page

pascalberger:
Additionally to commits to `cake-build` repositories, also return commits to `cake-contrib` repositories.

To have contributors which only contributed to cake-contrib repos also listed https://github.com/cake-contrib/Cake.AddinDiscoverer/issues/225 is required.


augustoproiete (approved):
LGTM

PR #2705, opened 2023-08-12
Update build to Cake 3.1

pascalberger:
Updates build to use Cake 3.1. 

Also updates the following addins:
- Cake.Wyam
- Cake.Yaml


augustoproiete (approved):
LGTM

PR #2711, opened 2023-08-14
Mention extension on the contributors page

pascalberger:
Since we're now also list contributors to cake-contrib org we should mention extensions in the introduction.


nils-a (approved):
LGTM!

PR #2712, opened 2023-08-14
List contributors on extension page

pascalberger:
List contributors on extension page for extensions hosted in the cake-contrib organization

image

### DIFF
--- a/input/_ExtensionsLayout.cshtml
+++ b/input/_ExtensionsLayout.cshtml
@@ -20,6 +20,28 @@
     string categories = (String.Join(" ", Model.List<string>("Categories").Select(x => $"<span class=\"badge badge-secondary\">{x}</span>")));
     var targetFrameworks = Model.List<string>("TargetFrameworks")?.Where(tfm => !string.IsNullOrWhiteSpace(tfm)).ToList() ?? new List<string>();
     var assemblies = Model.List<string>("Assemblies");
+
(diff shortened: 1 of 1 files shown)


input/_ExtensionsLayout.cshtml
```diff
@@ -158,11 +180,35 @@
             @author
         </p>
 
+        @if (contributors.Any())
+        {
+            <h2 class="no-anchor">Contributors</h2>
+            <p>
+                <ul class="list-unstyled">
+                    @foreach (var contributor in contributors)
```
nils-a, line 188: @pascalberger Should we remove the bots from the listing?

image

nils-a (approved):
One minor point (bots in the listing) - the rest LGTM!

input/_ExtensionsLayout.cshtml
pascalberger: @nils-a Data comes from [contributors.json](https://github.com/cake-build/website/blob/master/contributors/contributors.json) which is maintained by Addin Discoverer tool. It already filters accounts marked as bots by GitHub and we can additionally maintain a blacklist there (see https://github.com/cake-contrib/Cake.AddinDiscoverer/issues/226).

The screenshot is from my local dev environment and based on an earlier code base before we filtered bots. In the meantime bots were already removed (see https://github.com/cake-build/website/pull/2714).

PR #2904, opened 2024-03-26
Fix build

pascalberger:
Humanizer v3.0.0-beta.13 has changed signature of `EnumHumanizeExtension` to a generic class of `T : struct, Enum` so that code tries to call it instead of `DateHumanizeExtension`.


augustoproiete (approved):
LGTM

PR #3035, opened 2024-09-26
Fix broken Wyam link

pascalberger:
Since wyam.io is no longer available, we should link to the GitHub page.


augustoproiete (approved):
LGTM

PR #3081, opened 2024-11-13
Update mentioning of .NET

pascalberger:
Update mentioning of .NET to match latest Cake version and .NET wording


devlead (approved):
LGTM

PR #3140, opened 2025-01-10
Update getting-help.md

JPTAXLAW:


devlead:
Thank you for your PR. After reviewing it, we don't see how it adds value to our project and will therefor close this PR. If you disagree please provide more context or examples of its benefits.

PR #3237, opened 2025-08-13
Adds new blog post for Cake.Sdk 5.0.25225.53-beta

devlead:
Highlights:
- .NET 10 Preview 7 compatibility
- Dependency updates
- Analyzer warning fixes
- Improved SDK versioning in file-based apps.


Jericho:
FYI: executing the `dotnet new cakefile` like so:

image

results in the foling new file being created:
image

Notice the double `.cs`

devlead:
Ah 

> FYI: executing the `dotnet new cakefile` like so:
> 
> image
> results in the foling new file being created: image
> 
> Notice the double `.cs`

ah yes I'll fix,

devlead:
> Seems like the Cake SDK is looking for a file called Example.sln in a folder called src, neither of which exist on my machine.

The example project isn't created by default, however you can pass `--IncludeExampleProject` i.e.
```bash
dotnet new cakefile --name cake --IncludeExampleProject true
```

Jericho:
I followed your instructions to create the example project and now, when I execute `dotnet cake.cs`, the cake.cs is completely ignored:

image

I guess I'm confused what we are doing here. I was under the impression `dotnet cake.cs` would execute the content of this file without the need for a solution file. I'm probably not understanding the objective of the new Cake SDK.

Jericho:
I'm an idiot! I just opened cake.cs and noticed it's referencing example.sln, just like you said.

image

That's where the reference to this sln is coming from! I thought cake.cs would execute something trivial like `Information("Hello World");` or something. It completely went over my head that the sample cake.cs references a fictitious Example solution.

If I could make a suggestion:  either simplify the template cake.so it works immediately after `dotnet new cakefile --name cake` without the need for an example solution OR change the instructions in your blog post to let people know that Example.sln is necessary and case.cs will not work without it.

devlead:
@Jericho Concrete suggestions are welcome, I'm away from the keyboard a couple hours, so blocked until then. 

Up to now, I’ve done all the work in isolation based on my expectations of how templates should be used day to day, which might not align with the broader audience, so feedback is appreciated.

input/blog/2025-08-13-cake-sdk-net-preview-7-update.md
```diff
@@ -0,0 +1,75 @@
...
+```bash
+dotnet new cakefile --name cake
+```
+
+And run it with:
+
+```bash
+dotnet cake.cs
+```
+
```
Jericho, line 65: Here's a suggestion to make it clear that cake.cs will not run out-of-the-box. It assumes the presence of a solution:

```suggestion
PLEASE NOTE: this sample cake.cs file assumes that you have a file called 'example.sln' located in a folder called `src`. If you don't have such a solution file, you can create it like so:

```bash
dotnet new cakefile --name cake --IncludeExampleProject true --force
```
```

Jericho:
Fair enough. 

I made a suggestion on your blog post to add a note about the assumption made by cake.cs and how to overcome it if you don't already have `src/example.sln` on your machine.

Jericho:
I also opened a PR to add a trivial cake script template which does not depend on the presence of an example solution file.

PR #3396, opened 2026-05-11
Fix: Add type="html" to content tag in Atom feed

### LINKED ISSUE #2778: Atom feed doesn't handle HTML in content properly
In Atom feed (https://cakebuild.net/blog/feed/atom/) HTML tags in content is not handled properly.

As reported in https://twitter.com/DavidRGardiner/status/1726220941124936168

ghost:
Fixes #2778

The `<content>` tag in the Atom feed was missing the `type="html"` 
attribute as required by the W3C Atom spec.

Reference: https://validator.w3.org/feed/docs/atom.html#text


ghost:

PR #3397, opened 2026-05-15
Added logging documentation

Saisneha9:
Added a logging documentation page under fundamentals.

This documentation includes:

* built-in logging aliases
* Spectre.Console support for advanced console output

### DIFF
--- a/input/docs/fundamentals/logging.md
+++ b/input/docs/fundamentals/logging.md
@@ -1,0 +1,19 @@
+---
+Title: Logging
+---
+
+# Logging
+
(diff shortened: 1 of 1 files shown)


Saisneha9:

devlead (changes requested):
For this to be fully useful we should have some code examples, and explanation of Cake logging verbosities and how they're controlled.

devlead:
This was fixed by #3407, so we're going to close this in favor of it. Thanks for your contribution.

PR #3416, opened 2026-06-03
fix: change Headings to H2 for TOC semantic HTML

### LINKED ISSUE #1080: Change to H2 for TOC
We currently use H1 in our document pages to build the TOC on the right hand column. To follow semantic best practices we should only have one H1 on the page (the title) and use headings starting with H2 for structuring page content.

We would need to configure Wyam to pickup H2 for building the TOC

Himanshu-29-tech:
Fixes #1080

Currently Headings() defaults to H1 for the TOC. 
Changed to Headings(2) to follow semantic HTML best practices 
— a page should only have one H1 (the title).

### DIFF
--- a/config.wyam
+++ b/config.wyam
@@ -88,7 +88,7 @@ Pipelines.Add("RenderDsl",
     Documents("DslAliases"),
     Razor()
         .WithLayout("/_DslLayout.cshtml"),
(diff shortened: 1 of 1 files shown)


Himanshu-29-tech:

